PR #1, opened 2021-09-30
aligninging with evaluation group

debajyotidatta:
This PR simply aligns with the evaluation group. It is in the form of a notebook so that anyone else who wants to evaluate the models can do it directly.

Note: The metrics may not align with the evaluation group. So we will design our own metric in that case.


jason-fries:
LGTM!

PR #2, opened 2021-10-01
feat: implementing DDI Corpus prompts.

drugilsberg:
Prompts for entities.

Signed-off-by: Matteo Manica <drugilsberg@gmail.com>


prompts/ddi.py
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+    def _init_prompts(self):
+        """Initialize prompts."""
+        self._prompts = {}
+        self._metadata = {}
+
+
+# NOTE: implemented prompts
+def list_entity_template(
```
jason-fries, line 78: This is great!! Much more useful and generic that my implementation! Ideally we should start building a set of generic prompting functions like this (and for the relational/QA transformations) in a separate module to use across datasets/tasks.

jason-fries (approved):
Awesome job -- thanks for submitting the PR and cleaning up the entity prompting functions.

prompts/ddi.py
drugilsberg: Hey @jason-fries glad you found this useful. If you want I can also spend sometime packaging the whole folder as a python module and refactoring a bit.

PR #5, opened 2021-10-20
add n2c2 smoking status prompts

bwang482:
This script requires providing two arguments:
```
--indir : the directory stores the XML data files
--outdir : the output directory
```

Data downloading requires user registration and entering username and password on the n2c2 website, therefore can't be scripted in `init_dataset()`.

### DIFF
--- a/prompts/n2c2_smoking.py
+++ b/prompts/n2c2_smoking.py
@@ -17,6 +17,10 @@ _TRAINING_FILE = "smokers_surrogate_train_all_version2.xml"
 _TEST_FILE = "smokers_surrogate_test_all_groundtruth_version2.xml"
 
 def load_smoking_corpus(path):
+    """
+    Load the data split.
+    :param path: train/test data path
+    """
     logger.info(f"Loading n2c2 Smoking Status data from {path}")
     with open(path) as raw:
         file = raw.read().strip()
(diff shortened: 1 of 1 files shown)


prompts/n2c2_smoking.py
```diff
@@ -0,0 +1,188 @@
...
+        :param data_root: Root folder containing target dataset
+        """
+        self.data_root = Path(data_root).resolve()
+        self._init_dataset()
+        self._init_prompts()
+
+    def _init_dataset(self):
+        """
+        Prepare smoking status data.
+        Download data manually from https://portal.dbmi.hms.harvard.edu/projects/n2c2-nlp/", which requires requires user registration and DUA.
```
jason-fries, line 55: Yeah, I don't know how to get around the manual download aspect here for n2c2 datasets. I think the most generic solution is just having a `fpath` option in the dataset init where a user can point to a downloaded file.

prompts/n2c2_smoking.py
```diff
@@ -0,0 +1,188 @@
...
+# Create Prompts + Helper Functions
+#
+def classify(x):
+    """Simple prompt for classification.
+    Args:
+        x: a note from the smoking status corpus.
+    Returns:
+        the prompt.
+    """
+    tmpl = '"{}"\n'
```
jason-fries, line 82: Do you know how long (in tokens) the average document is here? We might have some issues with overly long prompts (>512 or >1024 tokens). For now I think it's fine, just something to be aware of.

bwang482: On average there are around 938 tokens per document (min. 148; **max. 3461**), using `medspacy_tokenizer`.

PR #7, opened 2021-11-04
Standardize the objects for prompting

hakunanatasha:
Creates a `Dataset` class that can process biomedical texts into a pybrat object for easy access to entities/relations.

Currently supports: jnlpba, cellfinder, linneaus, chemprot, ddi, bc5cdr by name

Custom-datasets are WIP; some initial templating holds for BioC and Brat can currently be parsed directly by pybrat.
Use case:
```
from dataloader import BioDataset

dataset = BioDataset("bc5cdr", "datasets") # Creates a directory "datasets" and downloads + processes
doc = dataset.data.train[0] # Single Document
text = doc.text # Document's text
Ent = dataset.data.train[0].entities[0] # Single Entity
Reln = dataset.data.train[0].relations[0] # Single Relation
```


leonweber:
I had a pass over all included data sets, visualised with https://gist.github.com/leonweber/fe7760ae24a4122e0104732e4213be39. Viewer can be used with `streamlit run viz.py`.

Downloading and parsing worked great for all data sets. This is really great work, thanks a lot @hakunanatasha !

*  "jnlpba":
    - Looks fine
* "cellfinder":
      - Looks fine
* "chemprot": 
      - Relation types should probably be human readable for prompting purposes ("regulator" instead of CPR:4). This should be automatically resolved if we use add DrugProt (https://biocreative.bioinformatics.udel.edu/tasks/biocreative-vii/track-1/) which supersedes ChemProt
* "linneaus":
       - Looks fine
* "ddi": DDI
      - Looks fine
* "bc5cdr": BC5CDR
       - Entity.type should probably be either "Drug" or "Chemical" and not the database identifier. Maybe we could just introduce our own entity class that has an additional field for data base ids.
       - CDR has no mention-level annotation. It seems like at the moment this is solved by connecting the last occurring mentions with the relation. This might be misleading if someone is not acquainted with the data set, because they might assume that this is an actual mention-level relation instead of a dummy. Maybe we should add additional entities which are not tied to a mention (Entity.start == Entity.end == None) and define the relations using those.

yongzx:
@hakunanatasha @leonweber 
This is the pipeline that you talked about in last Thursday's meeting update right?
Any way I can use it right away for setting up TANL?

Another question is that is it possible for you to aggregate the label types into a dict? An example for CoNLL03 is the following. For now, our key and value would be the same, but in the future, the value should be in natural language (i.e., we want to map labels to their natural-language labels).
```
    natural_entity_types = {
        'Loc': 'location',
        'Org': 'organization',
        'Peop': 'person',
        'Other': 'other',
    }

    natural_relation_types = {
        'Work_For': 'works for',
        'Kill': 'kills',
        'OrgBased_In': 'organization based in',
        'Live_In': 'lives in',
        'Located_In': 'located in'
    }
```

hakunanatasha:
@leonweber big thanks for reviewing - I'll try to get to this ASAP
@yongzx yes; but we haven't introduced the token-specific changes that TANL needs yet

hakunanatasha:
@yongzx seems like the inclusion of TANL is affecting the original script's ability to download and process; can you make a unit-test (or alternatively I can help) to check if the base example works?

hakunanatasha:
@yongzx I spoke too soon - looks like it's just bc5cdr which I know has some what of a strange set of entity start/ends due to discontinuous entities in text

jason-fries:
This looks awesome everyone!

PR #8, opened 2021-11-11
WIP: zero-shot and fewshot baselines

debajyotidatta:
Let's not merge until we have dataset specific numbers.


jason-fries:
@debajyotidatta Just cleaning up the hackathon version of the repo, so these few-shot tools are out-of-date (we moved the old main to a new branch). Post-hackathon we should revisit how to set these experiments up with the new infra :)

PR #76, opened 2022-01-28
added ParaMed Dataset for translation

ruisi-su:
The data directory already points to the remote HF datasets location. Let me know if I should remove the submodule and have them as copies of the same files hosted on HF datasets


ruisi-su:
updated PR with changes to comply with the t2t schema.

Issue #75, opened 2022-01-27
Create a dataset loader for MLEE


leonweber:
From http://www.nactem.ac.uk/MLEE/

PR #77, opened 2022-01-29
Add MLEE dataset

leonweber:
Addresses #75


leonweber:
@jason-fries referencing your review in #106:
* Naming of configs should now conform with guidelines
* KB schema should now conform with current version
* I agree that factoring the BRAT -> BigBio conversion out of this file would be a good idea. However I'm not sure where it could live. I was under the impression that the dataset scripts should be self contained, right? So another file would probably not work?

Issue #16, opened 2022-01-21
Create dataset loader for MuchMore


hakunanatasha:
From https://muchmore.dfki.de/resources1.htm

galtay:
hello, i would like to work on this dataset please.

PR #78, opened 2022-01-29
Add muchmore dataset

galtay:
The muchmore dataset supports multiple tasks and is in multiple languages. 

This PR implements multiple configs for different tasks and language combinations.  
This is a parallel corpus in english and german so there are a few configs that give the options of
* both
* english only
* german only

so far it supports, 
* original data format
* just plain text abstracts
* translation
* entity linking (spans -> groups of UMLS CUIs)


galtay:
connected to #16

galtay:
@ruisi-su added a config and format to match the paramed dataset you implemented.
https://huggingface.co/datasets/bigscience-biomedical/paramed/blob/main/paramed.py

 tried out the following, 
`config.name = "muchmore - translation"`
```
        elif self.config.name == "muchmore - translation":                                                   
            features = datasets.Features({                                                                   
                "en": datasets.Value("string"),                                                              
                "de": datasets.Value("string"),                                                              
                "metadata": {                                                                                
                    "sample_id_prefix": datasets.Value("string"),                                            
                    "sample_id_en": datasets.Value("string"),                                                
                    "sample_id_de": datasets.Value("string"),                                                
                }                                                                                            
            }) 
```

galtay:
will drop test file before merging, but just checking it in now to keep track.

galtay:
@hakunanatasha @jason-fries @sunnnymskang 
can we get some eyes on this PR? I think this is a good example for datasets that support more than one task

galtay:
@hakunanatasha @sunnnymskang @jason-fries @debajyotidatta @leonweber any thoughts on this one? this is an example of a dataset that supports multiple tasks.

PR #83, opened 2022-01-31
add n2c2_2011_coref

galtay:
First sketch of canonical coref schema using the n2c2 2011 coref task.

tl;dr doesn't have a coref schema yet! 

implemented the following, 
* reading from local files for data behind a data users agreement
* read out original form of data
* started sketching potential coref schema in docstring

i think the schema should have the roughly the following shape, 

* text
* entities (IDs, text offsets, type, other metadata)
* corefs (sets of entity IDs that are coreferent)

### DIFF
--- a/examples/n2c2_2011_coref.py
+++ b/examples/n2c2_2011_coref.py
@@ -1,0 +1,289 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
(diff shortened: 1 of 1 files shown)


examples/n2c2_2011_coref.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+class N2C22011CorefDataset(datasets.GeneratorBasedBuilder):
+    """n2c2 2011 coreference task"""
+
+    VERSION = datasets.Version(_VERSION)
+
+    BUILDER_CONFIGS = [
+        datasets.BuilderConfig(
+            name="original",
```
jason-fries, line 365: change to `source`

examples/n2c2_2011_coref.py
```diff
@@ -0,0 +1,588 @@
...
+            version=VERSION,
+            description="Original n2c2 2011 coreference schema",
+        ),
+        datasets.BuilderConfig(
+            name="bigbio-kb",
+            version=VERSION,
+            description="Simplified schema for BigScience-Biomedical KB tasks",
+        ),
+        datasets.BuilderConfig(
+            name="bigbio-kb-list",
```
jason-fries, line 375: I'm sold on list being the best choice here. If you want to make changes to support this as the default that would be excellent.

jason-fries (changes requested):
Great job @galtay ! n2c2 dataset formats are such a pain -- I'm excited to see them put into a sane format :) 
I think moving to the list schema and changing the default config name are the only big changes before merging.

galtay:
thanks for the review! made the schema name and seq -> list changes so going to merge this in.

PR #84, opened 2022-01-31
update template local to include data_dir and data_files kwargs

galtay:
Added some more details around the `data_dir` and `data_files` kwargs that can be used to pass path information into the `datasets.load_dataset` for datasets that are only available on the users hard drive. 

Also shorted up some long lines and did a bit of copy editing (not super opinionated on any of it so feel free to suggest edits, reversions, deletions, ...)

### DIFF
--- a/templates/template_local.py
+++ b/templates/template_local.py
@@ -14,31 +14,35 @@
 # limitations under the License.
 
 """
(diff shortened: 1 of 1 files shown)


hakunanatasha (approved):
LGTM - mostly doc comments but the builderconfig info was nice, thank for detailing out the comments more

PR #89, opened 2022-02-03
Biosses add

debajyotidatta:
Issue: Original Data Source link doesn't work with `dl_manager.download_and_extract`.

So instead, downloaded the data, processed it separately. Source now stored at examples folder.


sunnnymskang (commented):
@debajyotidatta @hakunanatasha Let's think through this together: Biosses dataset seems to be a bit tricky  1. There's no obvious matching to the current schema type (let me know if you think otherwise) due to additional annotator score columns (4 provided)  2. I believe @debajyotidatta had problem with pulling dataset using dl_manager (as per slack) and pushed the dataset to examples folder. On the other hand, in our hackathon guide, we're strongly recommend against pushing the dataset when they PR. I'm a bit torn about the current approach, as it might confuse the participants

debajyotidatta:
hi @sunnnymskang @hakunanatasha : It was decided in the meeting yesterday to just not add the dataset to the repo but to get it from any other repo. I have removed the dataset from the repository and it gets it using the `dl_manager` like all other datasets in the huggingface repository.

sunnnymskang:
@debajyotidatta Noted. Was there anything discussed about the schema type?

debajyotidatta:
Nothing. I think we can merge this. There will always be datasets with the format being slightly different. Do you want me to split this up into a `bigbio` format that gets rid of the user annotation just so we are consistent schema? This approach changes the original schema.

jason-fries:
The `bigbio` schema should not add any additional field names. By design, we're admitting that this schema might not represent the original task perfectly. We're favoring consistency here, but it's also the reason we additionally include the  `source` schema to support a non-destructive (or at least less destructive :P ) view of the dataset.

For the current PR, we should remove the `annotato_x` fields from `bigbio`
```
 "annotator_a": datasets.Value("int64"),
 "annotator_b": datasets.Value("int64"),
 "annotator_c": datasets.Value("int64"),
 "annotator_d": datasets.Value("int64"),
 "annotator_e": datasets.Value("int64")
```

sunnnymskang:
Sorry for missing on this thread - I echo with what Jason has suggested. I'll remove the additional annotators field from the bigbio schema, test if they work and will merge to master

sunnnymskang:
Merging to https://github.com/bigscience-workshop/biomedical/pull/178

PR #90, opened 2022-02-11
Biomedical Schema Draft

jason-fries:
- Please comment on the candidate task schemas. 
- Use PR review to add your thoughts and propose changes.

### DIFF
--- a/examples/bioasq9b.py
+++ b/examples/bioasq9b.py
@@ -110,6 +110,13 @@ class Bioasq9bDataset(datasets.GeneratorBasedBuilder):
                     "concepts": datasets.Sequence(datasets.Value("string")),
                     "ideal_answer": datasets.Sequence(datasets.Value("string")),
                     "exact_answer": datasets.Sequence(datasets.Value("string")),
+                    "triples": datasets.Sequence(
+                        {
+                            "p": datasets.Value("string"),
+                            "s": datasets.Value("string"),
+                            "o": datasets.Value("string"),
+                        }
+                    ),
                     "snippets": datasets.Sequence(
                         {
                             "offsetInBeginSection": datasets.Value("int32"),
@@ -127,15 +134,12 @@ class Bioasq9bDataset(datasets.GeneratorBasedBuilder):
             features = datasets.Features(
                 {
                     "id": datasets.Value("string"),
+                    "document_id": datasets.Value("string"),
+                    "question_id": datasets.Value("string"),
                     "question": datasets.Value("string"),
                     "type": datasets.Value("string"),
                     "context": datasets.Value("string"),
-                    "document": {
-                        "doc_id": datasets.Value("string"),
-                        "text": datasets.Value("string"),
-                    },
                     "answer": datasets.Sequence(datasets.Value("string")),
-                    "expanded_answer": datasets.Sequence(datasets.Value("string")),
                 }
             )
 
@@ -205,18 +209,24 @@ class Bioasq9bDataset(datasets.GeneratorBasedBuilder):
                         "id": record["id"],
                         "type": record["type"],
                         "body": record["body"],
+                        "documents": record["documents"],
+                        "concepts": record["concepts"] if "concepts" in record else [],
+                        "triples": record["triplets"] if "triples" in record else [],
+                        "ideal_answer": record["ideal_answer"],
+                        "exact_answer": record["exact_answer"],
+                        "snippets": record["snippets"],
                     }
+
                 elif self.config.name == "bigbio":
-                    # TBD: materialize each record as question X snippet/context
-                    yield record["id"], {
-                        "id": record["id"],
-                        "type": record["body"],
-                        "question": record["type"],
-                        "context": record["body"],
-                        "document": {
-                            "doc_id": "",
-                            "text": "",
-                        },
-                        "answer": record["body"],
-                        "expanded_answer": record["body"],
-                    }
+                    # instances are defined over snippets/documents
+                    for i, snippet in enumerate(record["snippets"]):
+                        uid = f'{record["id"]}_{i}'
+                        yield uid, {
+                            "id": record["id"],
+                            "document_id": snippet["document"],
+                            "question_id": record["id"],
+                            "question": record["body"],
+                            "type": record["type"],
+                            "context": snippet["text"],
+                            "answer": record["exact_answer"],
+                        }
--- a/examples/bioasq9b.py
+++ b/examples/bioasq9b.py
@@ -1,0 +1,222 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+BioASQ Task B On Biomedical Semantic QA (Involves IR, QA, Summarization qnd
+More). This task uses benchmark datasets containing development and test
+questions, in English, alongwith gold standard (reference) answers constructed
+by a team of biomedical experts. The participants have to respond with relevant
+concepts, articles, snippets and RDF triples, fromdesignated resources, as well
+as exact and 'ideal' answers.
+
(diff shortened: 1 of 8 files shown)


examples/bioasq9b.py
```diff
@@ -0,0 +1,232 @@
...
+            name="original",
+            version=VERSION,
+            description="Original BioASQ9 source schema.",
+        ),
+        datasets.BuilderConfig(
+            name="bigbio",
+            version=VERSION,
+            description="Simplified schema for BigScience-Biomedical QA tasks",
+        ),
+    ]
```
galtay, line 96: I like the `original` and `bigbio` config names ... we might want to do something like `bigbio-schema` (e.g. bigbio-kb or bigbio-text-pairs) as some datasets support multiple tasks. Specifically I'm thinking of things like the MuchMore dataset which can be a translation dataset or an NER dataset. `bigbio-<schema>` instead of `bigbio-task` or just `bigbio` handles this well i think

examples/bioasq9b.py
```diff
@@ -0,0 +1,232 @@
...
+                            "text": datasets.Value("string"),
+                            "beginSection": datasets.Value("string"),
+                            "endSection": datasets.Value("string"),
+                            "document": datasets.Value("string"),
+                        }
+                    ),
+                }
+            )
+        # simplified scheam for QA tasks
+        elif self.config.name == "bigbio":
```
galtay, line 133: e.g. `bigbio-qa` here instead of `bigbio` ... this would also allow some nice discovery by investigating the configs of the datasets

examples/bioasq9b.py
```diff
@@ -0,0 +1,232 @@
...
+                name=datasets.Split.TEST,
+                # These kwargs will be passed to _generate_examples
+                gen_kwargs={
+                    "filepath": os.path.join(
+                        data_dir, "BioASQ-training9b/training9b.json"
+                    ),
+                    "split": "test",
+                },
+            ),
+        ]
```
galtay, line 193: we should probably leave some instructions in the hackathon README about what to do if the dataset does not contain a canonical train/val/test split. looks like these files are the same for train and test ... I think we should default to just a `train` set if there is no canonical split.

examples/bioasq9b.py
```diff
@@ -0,0 +1,232 @@
...
+    ):
+        """Yields examples as (key, example) tuples."""
+        # This method handles input defined in _split_generators to yield (key, example) tuples from the dataset.
+        # The `key` is here for legacy reason (tfds) and is not important in itself.
+
+        with open(filepath, encoding="utf-8") as file:
+            data = json.load(file)
+            for record in enumerate(data):
+                if self.config.name == "original":
+                    yield record["id"], {
```
galtay, line 208: I might be confused, but I think `record` here will be a tuple with the first element i = 0 to len(data)-1 and the second element being data[i]  .. so `record["id"]` seems like it would throw an error. I think maybe we want something like, 
```
for _id, record in enumerate(data):
    yield _id, {"id": record["id"], ...}
```

schemas/entailment.py
```diff
@@ -0,0 +1,14 @@
...
+"""
+import datasets
+
+
+features = datasets.Features(
+    {
+        "id": datasets.Value("string"),
+        "premise": datasets.Value("string"),
+        "hypothesis": datasets.Value("string"),
+        "label": datasets.Value("string"),
```
galtay, line 12: add even more guide rails and make this 
`"label": datasets.ClassLabel(names=["entailment", "contradiction", "neutral"])`
or too strict?

schemas/kb.py
```diff
@@ -0,0 +1,81 @@
...
+and a flat hierarchy of document passages.
+
+"""
+import datasets
+
+
+features = datasets.Features(
+    {
+        "id": datasets.Value("string"),
+        "document_id": datasets.Value("string"),
```
galtay, line 23: I think it would be nice to have a top level `text` field here that any offset would work in ... for example
```
text = sample["text"]
ii = sample["passages"][n]["offsets"][0][0]
ff = sample["passages"][n]["offsets"][0][1]
assert(text[ii:ff] = sample["passages"][n]["text"])
# same for entities 
```

Otherwise, if I'm looking for the context of an entity I have to search through each passage to discover if the entity offset is within the offset range of that passage. This could be as straightforward as, 
```
sample["text"] = " ".join([passage['text'] for passage in passages])
```

It definitely leads to some duplicate information but I think its worth it for reasonable sized datasets ... it would also give the unit tests something to check.

task_schemas.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Named entity recognition (NER)
+- Named entity disambiguation/normalization/linking (NED)
+- Event extraction
+- Relation extraction (RE)
+- Coreference resolution
+
+```
+{
+    "document_id": "XXXXXX",
```
galtay, line 31: we have `id` and `document_id` in the `schemas/kb.py`  maybe we can go into some detail here about how they should be used?

task_schemas.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+### Passages
+
+Passages capture document structure, such as the title and abstact sections of a PubMed abstract. 
+
+```
+{
+    "id": 0,
+    "document_id": "227508",
```
galtay, line 60: this makes the distinction between `id` and `document_id` a bit more clear .... do we want `id` to be something like "a unique hashable item in the dataset"?

examples/bioasq9b.py
```diff
@@ -0,0 +1,232 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors and the current dataset script contributor.
```
galtay, line 2: do we want to have people add their name here in place of `the current dataset script contributor`? or have people append their name if they make a PR?

schemas/kb.py
galtay: coming back to this (sorry for the long comments!) ... I now see that my suggestion would be particularly silly for samples with one passage (then the top level text would just be an exact copy of the `passages[0]['text']`  ... I withdraw my top level text suggestion :)

task_schemas.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+**Schema Notes**
+
+- `id` must be unique string within dataset and across parent category (e.g., entities, passages)
+- `offsets` contains absolute character offsets per span
+- `offsets` and `text` are lists to support discontinous spans. 
+- `normalized` may contain 1 or more normalized links to database entity identifiers.
+- `passages` captures document structure such as named sections.
+- `entities`,`events`,`coreferences`,`relations` may be empty fields based on dataset and specific task.
```
galtay, line 49: we might even go as far as to leave these fields out if they are not included in the dataset?

schemas/kb.py
```diff
@@ -0,0 +1,81 @@
...
+                "type": datasets.Value("string"),
+                "text": datasets.Value("string"),
+                "offsets": datasets.Sequence([datasets.Value("int32")]),
+            }
+        ),
+        "entities": datasets.Sequence(
+            {
+                "id": datasets.Value("string"),
+                "offsets": datasets.Sequence([datasets.Value("int32")]),
+                "text": datasets.Sequence([datasets.Value("string")]),
```
galtay, line 36: I think this should be just `Sequence(Value("string"))` not a nested iterable

schemas/kb.py
```diff
@@ -0,0 +1,81 @@
...
+features = datasets.Features(
+    {
+        "id": datasets.Value("string"),
+        "document_id": datasets.Value("string"),
+        "passages": datasets.Sequence(
+            {
+                "id": datasets.Value("string"),
+                "type": datasets.Value("string"),
+                "text": datasets.Value("string"),
+                "offsets": datasets.Sequence([datasets.Value("int32")]),
```
galtay, line 29: if we're not allowing for disjoint passages I think we should have just `Sequence(Value("int32"))` here  (OR if we are, then make the `text` a sequence as well). I think we don't need disjoint passages.

examples/bioasq9b.py
jason-fries: This is just an oversight in this case -- there is a canonical test set. More generally, I think we should be in a setting where there is always a train split defined and (hopefully) a public test or validation set. If not, defaulting to `train` is good with me

schemas/entailment.py
jason-fries: Yeah, I'm torn. Let's leave this unrestricted for now and see if it becomes a problem and we can flag this for fixing in a post-hackathon cleanup pass.

schemas/kb.py
jason-fries: This is to match the discontinuous offset definition. The ugliness of supporting discontinuous spans necessitates this unfortunately.

galtay: Even for the discontinuous case, a nested set of offsets `[(0,4), (9,12)]` leads to one less level of nesting for the text right? `["text1", "text2"]`

PR #102, opened 2022-02-15
Adding more details to test your data loader

galtay:


UPLOADING.md
```diff
@@ -65,10 +65,31 @@ At this point, there should be **no further changes in your code**.
 
 Run the following command **in a folder that does not include your data-loading script**:
 
+Test both the original dataset schema/config and the bigbio schema/config. 
+
+```python
+from datasets import load_dataset
+
+dataset_orig = load_dataset("bigscience-biomedical/<your_dataset_name>", name="source", use_auth_token=True)
```
sunnnymskang, line 73: @galtay Can you provide an example/context at which users would need to pull the dataset under different names?

UPLOADING.md
```diff
@@ -65,10 +65,31 @@ At this point, there should be **no further changes in your code**.
...
+dataset_bigbio = load_dataset("bigscience-biomedical/<your_dataset_name>", name="bigbio", use_auth_token=True)
+```
+
+For datasets that are licensed such that they must already exist on the users local machine, 
+the test commands might look something like this instead,
```
sunnnymskang, line 78: @galtay Can you provide an example of the case where "datasets that are licensed such that they must already exist on the users local machine"

UPLOADING.md
galtay: `name` is how you specify the particular dataset loader config you want to use. for the hack-a-thon we ask participants to implement the original dataset schema (i.e. as close as we can get to the native dataset format) as a `source` config and then to also implement the `bigbio` config for the `bigbio` schema relevant to the dataset task (e.g. `kb`, `translation`, ...) . All of the implementations in the `examples` folder should do this ... this one is probably easiest to view at the moment https://github.com/bigscience-workshop/biomedical/blob/master/examples/bc5cdr.py#L80

UPLOADING.md
galtay: here is an example of a dataset that cannot be publicly downloaded but we have a dataset loader script that the user can use and point to their local copy of data. https://github.com/bigscience-workshop/biomedical/blob/master/examples/n2c2_2011_coref.py ... In the README FAQs this is is covered under the 
```
What if my dataset does not have a public license?

We understand that some biomedical datasets require external licensing. To respect the agreement of the license, we recommend implementing a dataloader script that works if the user has a locally downloaded file
```
eventually we should make sure that this part of the FAQs points to an up to date, but also simple example and template and make clear how the use of `data_dir` and `data_files` keyword arguments are supported in the `datasets` package. 
See the `Note` in the `datasets` docs for a bit more background https://huggingface.co/docs/datasets/add_dataset.html#downloading-data-files-and-organizing-splits

UPLOADING.md
hakunanatasha: im going to add this because its easier to make edits on top of what's here - i approve these changes because the authtoken and the name of the view are useful commands to default to. we will need to add more info around how to access views etc and what they might mean but fine for now

hakunanatasha (approved):
changes to dataloader so that it uses a view-specific load

PR #104, opened 2022-02-15
text/offset iterables update

galtay:
The text field should have one less level of nesting than the offsets.
For passages, 
offsets = (lo,hi) and text = str 
For entities,
offsets = [(lo,hi), (lo,hi)] and text = ["txt1", "txt2"]

Sequence([...]) is an iterable inside an iterable. Sequence() is just an iterable.


jason-fries (approved):
Lgtm!

PR #105, opened 2022-02-16
Schema examples for BC5CDR and BioASQ9b updated

jason-fries:
This fully implements BC5CDR and BioASQ9b in the source (original) and bigbio schemas as worked examples for the hackathon.


examples/bc5cdr.py
```diff
@@ -58,65 +65,110 @@
...
+            features = datasets.Features(
+                {
+                    "id": datasets.Value("string"),
+                    "document_id": datasets.Value("string"),
+                    "passages": [
+                        {
+                            "id": datasets.Value("string"),
+                            "type": datasets.Value("string"),
+                            "text": datasets.Value("string"),
+                            "offsets": datasets.Sequence([datasets.Value("int32")]),
```
galtay, line 141:
```suggestion
                            "offsets": datasets.Sequence(datasets.Value("int32")),
```

examples/bc5cdr.py
```diff
@@ -183,45 +227,99 @@ def _split_generators(self, dl_manager):
...
+                    "passages": [],
+                    "entities": [],
+                    "relations": [],
+                }
+                uid += 1
+
+                char_start = 0
+                # passages must not overlap and spans must cover the entire document
+                for passage in xdoc.passages:
+                    offsets = [[char_start, char_start + len(passage.text)]]
```
galtay, line 293: don't need nested iterables for passage offsets and its incompatible with the single string field for `text`

galtay (changes requested):
Nice example! Looks like there is just one offset iterable change to make and then LGTM

examples/bc5cdr.py
jason-fries: Using`datasets.Sequence([datasets.Value("int32")])` for consistency with entities, etc, per slack conversation

examples/bc5cdr.py
jason-fries: Again. matching for consistency per Slack

PR #106, opened 2022-02-16
KB Schema Update: Add support for event triggers

leonweber:
This introduces support for event triggers. In the old schema "trigger" was a reference but there wasn't any field to put the referenced text span.

### DIFF
--- a/examples/mlee.py
+++ b/examples/mlee.py
@@ -1,0 +1,426 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This is a template on how to implement a dataset in the biomedical repo.
+
+A thorough walkthrough on how to implement a dataset can be found here:
+https://huggingface.co/docs/datasets/add_dataset.html
+
+This script corresponds to Step 4 in the Biomedical Hackathon guide.
+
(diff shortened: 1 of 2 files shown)


examples/mlee.py
```diff
@@ -0,0 +1,573 @@
...
+- `_DATASETNAME` = "your_dataset_name"
+- `_CITATION`: Latex-style citation of the dataset
+- `_DESCRIPTION`: Explanation of the dataset
+- `_HOMEPAGE`: Where to find the dataset's hosted location
+- `_LICENSE`: License to use the dataset
+- `_URLs`: How to download the dataset(s), by name; make this in the form of a dictionary where <dataset_name> is the key and <url_of_dataset> is the value
+- `_VERSION`: Version of the dataset
+"""
+
+_DATASETNAME = "mlee"
```
jason-fries, line 63: We're running with "source" as the name of the original schema and adding a suffix if there are multiple tasks in the dataset (e.g., `source-qa`).
For the bigscience schema, we've just been going with bigbio :)

examples/mlee.py
```diff
@@ -0,0 +1,573 @@
...
+                            ),  # ID in the resource, e.g. 534366
+                            "text": datasets.Value(
+                                "string"
+                            ),  # Human readable description/name of the entity, e.g. "Barack Obama"
+                        }
+                    ),
+                },
+            )
+        elif self.config.name == _UNIFIED_VIEW_NAME:
+            features = datasets.Features(
```
jason-fries, line 245: This should match the schema format finalized in `schemas/kb.py` -- looks like there are some mismatches here.

examples/mlee.py
```diff
@@ -0,0 +1,573 @@
...
+                {
+                    "head": ann["head"]["ref_id"],
+                    "tail": ann["tail"]["ref_id"],
+                }
+            )
+
+        return unified_example
+
+
+    def parse_brat_file(self, txt_file):
```
jason-fries, line 453: We might want to move this out to a separate helper module since it looks pretty intimidating for an example file! We could also consider using a standalone [BRAT Standoff file module](https://github.com/snorkel-team/snorkel-extraction/blob/ce9551bd47f1f0a8e52fa88c155e3bb43f636b8b/snorkel/contrib/brat/brat.py#L495).

jason-fries (changes requested):
Agree with the event trigger schema changes. Let's update the schema example to match the final format and then we can merge.

leonweber:
Oh my bad, the MLEE addition was not supposed to be included in this PR. I opened a separate PR for this #77
The event trigger changes will be resolved via #109, so I'm closing this one.

PR #107, opened 2022-02-16
Update CONTRIBUTING.md

bwang482:


CONTRIBUTING.md
```diff
@@ -10,7 +10,7 @@ Fork the dataset [repository](https://github.com/huggingface/datasets) from hugg
 
 After you fork, clone the repository locally. You can do so as follows:
 
-    git clone git@github.com:<your_github_username>/datasets.git
+    git clone https://github.com/<your Github handle>/datasets
```
galtay, line 13: @bwang482 do you specifically want people to use https as opposed to ssh?

bwang482: No, but below was the error before changing to https. https is used in  https://github.com/huggingface/datasets/blob/master/ADD_NEW_DATASET.md#start-by-preparing-your-environment 

```
git clone git@github.com:bwang482/datasets.git
Cloning into 'datasets'...
The authenticity of host 'github.com (140.82.112.3)' can't be established.
ED25519 key fingerprint is SHA256:+DiY3wvvV6TuJJhbpZisF/zLDA0zPMSvHdkr4UvCOqU.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

hakunanatasha: @bwang482 this will be deprecated because we no longer need to fork `datasets`; let's see if this issue arises with biomed

PR #108, opened 2022-02-16
n2c2_2011_coref update offset/text nesting, source + bigbio version

galtay:


jason-fries (changes requested):
I think this looks good except perhaps removing `"bigbio-sequence"` schema option and confirming that `bigbio` matches the final schema.

galtay:
going to merge this, but we can revisit if new issues come up.

PR #121, opened 2022-02-18
Add tests for bigbio KB datasets

leonweber:
This adds a few tests for the bigbio KB datasets, which can be called like this:
```bash
python tests/test_kb_dataset.py [dataset_script]
```
* Can datasets be created from `source` and `bigbio`. `datasets` automatically casts values to the type specified in `features`  so we don't need to check for that explicitly.
* Print counts for all schema elements for quick sanity checks
* Test whether the used `features` are *identical* to `schemas/kb.py`. We have to discuss whether we really want identity or whether the used schema can also be a subset of `schemas/kb.py`
* Test whether all referenced ids exist and have the correct type (e.g. entity or event)
* Test whether ids are globally unique

### DIFF
--- a/tests/test_kb_dataset.py
+++ b/tests/test_kb_dataset.py
@@ -2,6 +2,7 @@ import argparse
 import sys
 import unittest
 from collections import defaultdict
+from collections.abc import Iterable
 from difflib import ndiff
 from pathlib import Path
 from pprint import pformat
@@ -14,13 +15,10 @@ import datasets
 from schemas.kb import features
 
 OFFSET_ERROR_MSG = (
-    "\n"
+    "\n\n"
     "There are features with wrong offsets!"
-    "This is not a hard failure, as it is common for this type of datasets."
-    "However, if the error list is long (e.g. >10) you should double check your code."
-    # I do not see an easy way to cacth this in the test
-    "You can safely ignore this error if the dataset contains composite annotations"
-    "(see entities in `examples/bc5cdr.py` as an example)."
+    " This is not a hard failure, as it is common for this type of datasets."
+    " However, if the error list is long (e.g. >10) you should double check your code."
 )
 
 
@@ -29,27 +27,6 @@ def _get_example_text(example: dict) -> str:
     Get full text example
     """
     return " ".join([t for p in example["passages"] for t in p["text"]])
-
-
-def _check_offsets(
-    example_text: str,
-    offsets: list[list[int]],
-    texts: list[str],
-) -> Iterator:
-
-    for idx, (start, end) in enumerate(offsets):
-
-        try:
-            text = texts[idx]
-
-        # TODO: riase error?
-        except IndexError:
-            print("Pairing of a list of offsets with a single text is not allowed!")
-
-        by_offset_text = example_text[start:end]
-
-        if by_offset_text != text:
-            yield f" text:`{text}` != offsets_text:`{by_offset_text}`"
 
 
 class TestKBDataset(unittest.TestCase):
@@ -201,10 +178,62 @@ class TestKBDataset(unittest.TestCase):
                     example_text = _get_example_text(example)
 
                     for passage in example["passages"]:
-                        for idx, (start, end) in enumerate(passage["offsets"]):
-                            self.assertEqual(
-                                example_text[start:end], passage["text"][idx]
-                            )
+
+                        text = passage["text"]
+                        offsets = passage["offsets"]
+
+                        with self.subTest(
+                            "Offsets in passages must have only one element",
+                            offsets=offsets,
+                        ):
+                            self.assertEqual(len(offsets), 1)
+
+                        with self.subTest(
+                            "Text in passages must be an Iterable",
+                            text=text,
+                        ):
+                            self.assertIsInstance(text, Iterable)
+
+                        with self.subTest(
+                            "Text in passages must have only one element",
+                            text=text,
+                        ):
+                            self.assertEqual(len(text), 1)
+
+                        for idx, (start, end) in enumerate(offsets):
+                            self.assertEqual(example_text[start:end], text[idx])
+
+    def _check_offsets(
+        self,
+        example_text: str,
+        offsets: list[list[int]],
+        texts: list[str],
+    ) -> Iterator:
+
+        with self.subTest(
+            "# of texts must be equal to # of offsets", texts=texts, offsets=offsets
+        ):
+            self.assertEqual(len(texts), len(offsets))
+
+        with self.subTest(
+            "Text fields paired with offsets must be in the form [`text`, ...]",
+            texts=texts,
+        ):
+            self.assertIsInstance(texts, Iterable)
+
+        with self.subTest(
+            "All offsets must be in the form [(lo1, hi1), ...]", offsets=offsets
+        ):
+            self.assertTrue(all(len(o) == 2 for o in offsets))
+
+        # offsets are always list of lists
+        for idx, (start, end) in enumerate(offsets):
+
+            by_offset_text = example_text[start:end]
+            text = texts[idx]
+
+            if by_offset_text != text:
+                yield f" text:`{text}` != text_by_offset:`{by_offset_text}`"
 
     def test_entities_offsets(self):
         """
@@ -225,7 +254,7 @@ class TestKBDataset(unittest.TestCase):
 
                     for entity in example["entities"]:
 
-                        for msg in _check_offsets(
+                        for msg in self._check_offsets(
                             example_text=example_text,
                             offsets=entity["offsets"],
                             texts=entity["text"],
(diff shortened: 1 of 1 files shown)


tests/test_kb_dataset.py
```diff
@@ -0,0 +1,127 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("datset_script")
+    args = parser.parse_args()
+
+    schema_file = Path(__file__).parent.parent / "schemas" / "kb.py"
+    # Load the schema
+    with open(schema_file) as f:
+        exec(f.read())
```
galtay, line 126: can we make this an import somehow?

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,127 @@
...
+
+        for event in example["events"]:
+            for argument in event["arguments"]:
+                referenced_ids.append((argument["ref_id"], "event_arg"))
+
+        for coreference in example["coreferences"]:
+            for entity_id in coreference["entity_ids"]:
+                referenced_ids.append((entity_id, "entity"))
+
+        for relation in example["relations"]:
```
galtay, line 100: if we are not requiring all fields, we can still do something like `example.get("relations", [])` here

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def runTest(self):
+        self.print_statistics()
+        self.test_is_bigbio_schema_compatible()
+        self.test_are_ids_globally_unique()
+        self.test_do_all_referenced_ids_exist()
+
+    def test_is_bigbio_schema_compatible(self):
+        for split in self.dataset_bigbio.values():
+            self.assertEqual(split.features, features)
```
galtay, line 44: My initial reaction was to not make all the features required, but I can see the simplicity of things like this if we do make everything required. I'm on the fence, but probably still slightly leaning towards not making all fields required.

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,127 @@
...
+            print("=" * 10)
+
+            counter = defaultdict(int)
+            for example in split:
+                for feature_name, feature in features.items():
+                    if isinstance(feature, datasets.Value) or isinstance(
+                        feature_name, dict
+                    ):
+                        if example[feature_name]:
+                            counter[feature_name] += 1
```
galtay, line 28: If I'm understanding this conditional correctly, I'd remove it. I think we can allow for some top level keys being missing if the data doesn't support it (i.e. no coreferences if they are not in the data). but i think we should enforce that every example has a given key if that key is included in the dataset. (i might be mis-reading these `for` loops though)

tests/test_kb_dataset.py
leonweber: yep, addressed with commit [0613907](https://github.com/bigscience-workshop/biomedical/pull/121/commits/0613907bcebe3c4ecc7f9295b89fbaa0d192b7d9)

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,310 @@
...
+    "This is not a hard failure, as it is common for this type of datasets"
+    "However, if the error list is long (e.g. >10) you should double check your code"
+)
+
+
+def _get_example_text(example: dict) -> str:
+    """
+    Get full text example
+    """
+    return " ".join([t for p in example["passages"] for t in p["text"]])
```
galtay, line 28: I think I prefer `"".join` as opposed to `" ".join` ... this way passages is a partition of the text with no gaps and the offsets for the passages wont have any gaps ... we could also add a test for that .. something like, 
```
example['passages'][n]['offsets'][0][1] = example['passages'][n-1]['offsets'][0][0] + 1
```
what do you think?

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,310 @@
...
+) -> Iterator:
+
+    for idx, (start, end) in enumerate(offsets):
+
+        try:
+            text = texts[idx]
+
+        # TODO: riase error?
+        except IndexError:
+            print("Pairing of a list of offsets with a single text is not allowed!")
```
galtay, line 44: I think we finally landed on treating all offsets and text the same way in the schema so we can probably pull this out to its own test that asserts, 
* all offsets are of the form `[(lo1, hi1), ...]`
* all text fields that are paired with offsets are of the form `[text, ..]`
* offsets in passages have only one element `[(lo1,hi1)]`
* texts in passages have only one element `[text]`

no strong feelings about raising error or how its presented to the user, just saying we can treat all offsets the same.

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,310 @@
...
+        self.dataset_source = datasets.load_dataset(
+            self.PATH,
+            name="source",
+            data_dir=self.DATA_DIR,
+            use_auth_token=self.USE_AUTH_TOKEN,
+        )
+
+        self.dataset_bigbio = datasets.load_dataset(
+            self.PATH,
+            name="bigbio",
```
galtay, line 69: probably want to allow a little bit of flexibility here ... every once in a while we'll have one dataset that supports multiple tasks and so might have  `bigbio-translation` and `bigbio-kb` in one dataset.

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,310 @@
...
+
+        for split in self.dataset_bigbio:
+
+            if "passages" in self.dataset_bigbio[split].features:
+
+                for example in self.dataset_bigbio[split]:
+
+                    example_text = _get_example_text(example)
+
+                    for entity in example["passages"]:
```
galtay, line 199: can we change the variable name from `entity` to `passage` ?

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,310 @@
...
+                            texts=entity["text"],
+                        ):
+
+                            entity_id = entity["id"]
+                            errors.append(
+                                f"Example:{example_id} - entity:{entity_id} " + msg
+                            )
+
+        if len(errors) > 0:
+            self.fail(msg="\n".join(errors) + OFFSET_ERROR_MSG)
```
galtay, line 236: I really like the idea of collecting some examples to show to the user .. very cool!

tests/test_kb_dataset.py
galtay: ok got it, yea, that makes sense. fine with leaving for now.

tests/test_kb_dataset.py
sg-wbi: I used the whitespace to join passages because from my experience this is sort of a convention of datasets in PubTator and BioC format. We can make "passages" contiguous but I think we need to say it explicitly somewhere then, as it might be in the "muscle memory" of people like me. 

Thumbs up on the test anyway. I will add as soon as we agree on this.

galtay: sounds good. if its more of a convention in bio datasets let's keep as `" "`

tests/test_kb_dataset.py
sg-wbi: My opinion is that adding tests for this would be helpful. This way we ensure the format. 

When I wrote this I was thinking more of the situation of an entity which has multiple offsets, e.g. should the format be

"offsets" : [(lo1, hi1), (lo1, hi1),...] , "text" : ["text"]

or do we ask to duplicate the text?

"offsets" : [(lo1, hi1), (lo1, hi1),...] , "text" : ["text", "text"]

tests/test_kb_dataset.py
sg-wbi: I'll add an argument for that!

tests/test_kb_dataset.py
sg-wbi: Thanks! People are not asked to fix the offsets but this turns useful in case they wanted to.

tests/test_kb_dataset.py
sg-wbi: Sure!

tests/test_kb_dataset.py
hakunanatasha: Vastly prefer the `get` call for edge cases; is there a reason we pass empty list instead of None for "invalid keys"? because I would opt strongly for nones unless you need some type behavior

galtay: `None` is not iterable so if you use `None` you have to write some code for both cases right? by returning an empty list if the component is not there, you can still iterate over it, but nothing will happen b/c its empty. in other words, you can write these for loops for every sub-component in the schema and it wont break if they are not there.

tests/test_kb_dataset.py
sg-wbi: > len(offsets) == len(text)
Is this true? What about entities with same text on multiple locations?

tests/test_kb_dataset.py
```diff
@@ -0,0 +1,350 @@
...
+                            "Offsets in passages must have only one element",
+                            offsets=offsets,
+                        ):
+                            self.assertEqual(len(offsets), 1)
+
+                        with self.subTest(
+                            "Text in passages must be an Iterable",
+                            text=text,
+                        ):
+                            self.assertIsInstance(text, Iterable)
```
galtay, line 195: a string is also an iterable :) but I think this suite of tests will catch all but the most pathological cases and will be a great tool for contributors

sg-wbi: > a string is also an iterable :)

😅 brain fart

PR #122, opened 2022-02-18
Add event extraction example (MLEE)

leonweber:
Implements event extraction example in form of the MLEE corpus. The PR was already reviewed in #106 and was updated to incorporate all comments but one. The remaining TODO is to move the brat parsing and the brat -> bigbio conversion into a utility folder/file as discussed in slack. I'm waiting with the refactoring until it's clear how the file structure for utilities will look like.


examples/mlee.py
```diff
@@ -0,0 +1,610 @@
...
+
+(2) `_split_generators`: Download and extract data for each split of the data (ex: train/dev/test)
+
+(3) `_generate_examples`: From downloaded + extracted data, process files for the data in a feature format specified in "info".
+
+----------------------
+Step 1: Declare imports
+Your imports go here; the only mandatory one is `datasets`, as methods and attributes from this library will be used throughout the script.
+
+We have provided some import statements that we strongly recommend. Feel free to adapt; so long as the style-guide requirements are satisfied (Step 5), then you should be able to push your code.
```
galtay, line 38: For the examples I think we should remove this boilerplate and put a description of the dataset (following https://github.com/huggingface/datasets/blob/master/templates/new_dataset_script.py)

examples/mlee.py
```diff
@@ -0,0 +1,610 @@
...
+
+    def _info(self):
+
+        if self.config.name == _SOURCE_VIEW_NAME:
+            features = datasets.Features(
+                {
+                    "id": datasets.Value("string"),
+                    "document_id": datasets.Value("string"),
+                    "text": datasets.Value("string"),
+                    "text_bound_annotations": datasets.Sequence(  # T line in brat, e.g. type or event trigger
```
galtay, line 168: I think we are using list of dict as opposed to sequence of dict now (same comment for other element below)

examples/mlee.py
```diff
@@ -0,0 +1,610 @@
...
+        Nested zip files can be cached also, but make sure to save their path.
+
+        Fill the arguments of "SplitGenerator" with `name` and `gen_kwargs`.
+
+        Note:
+
+        - `name` can be: datasets.Split.<TRAIN/TEST/VALIDATION> or a string
+        - all keys in `gen_kwargs` can be passed to `_generate_examples()`. If your dataset has multiple files, you can make a separate key for each file, as shown below:
+
+        """
```
galtay, line 345: I think we can get rid of most of this boilerplate too and optionally replace with example specific comments ?

examples/mlee.py
```diff
@@ -0,0 +1,610 @@
...
+
+        The goal of this function is to perform operations on any of the keys of `gen_kwargs` that allow you to extract and process the data.
+
+        The following skeleton does the following:
+
+        - "extracts" abstracts
+        - "extracts" entities, assuming the output is of the form specified in `_info`
+        - "extracts" relations, assuming similarly the output in the form specified in `_info`.
+
+        An assumption in this pseudo code is that the abstract, entity, and relation file all have linking keys.
```
galtay, line 388: Here too, I think the boilerplate here is actively misleading for a participant.

galtay (commented):
Just a couple comments, but looks good.

examples/mlee.py
leonweber: Thanks for the review! The boilerplate comments were indeed misleading. I have replaced them with comments specific to MLEE which are hopefully more useful in 6525b94fea387d4fc0346e90f056c507313a393a.

examples/mlee.py
leonweber: Resolved via 6525b94fea387d4fc0346e90f056c507313a393a

examples/mlee.py
leonweber: Resolved via 6525b94fea387d4fc0346e90f056c507313a393a

examples/mlee.py
leonweber: Changed for all occurrences in bb2c660abe44819de088b5ec375511f00838ee4e

galtay-tempus (approved):

PR #123, opened 2022-02-18
test cases for text in passage/entity/events

sg-wbi:
test_kb.py [-h] --path PATH [--data_dir DATA_DIR] [--use_auth_token USE_AUTH_TOKEN]

WHAT's in here:
- tests for the passage/entities/events text fields, i.e. check that the string reported in the text field is the same as the one extracted by the offsets
- tests are skipped if the feature is not found in the split `feature` attribute (schema subset)
- tests for entities and events offset print a message re-assuring that this is not too bad, i.e. many datasets have wrong offsets, but if the there are many errors the code should be checked again
- verify all entity ids referenced by `coreferences` are there

### DIFF
--- a/tests/test_kb.py
+++ b/tests/test_kb.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Tests for KB template
+"""
+
+import argparse
+import sys
+import unittest
+from typing import Iterator, Optional, Union
+
+from datasets import load_dataset
+
+OFFSET_ERROR_MSG = (
+    "\n"
+    "There are features with wrong offsets!"
+    "This is not a hard failure, as it is common for this type of datasets"
+    "However, if the error list is long (e.g. >10) you should double check your code"
+)
+
+
+def _get_example_text(example: dict) -> str:
+    """
+    Get full text example
+    """
+    return " ".join([t for p in example["passages"] for t in p["text"]])
+
+
+def _check_offsets(
+    example_text: str,
+    offsets: list[list[int]],
+    texts: list[str],
+) -> Iterator:
+
+    for idx, (start, end) in enumerate(offsets):
+
+        try:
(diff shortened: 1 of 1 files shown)


tests/test_kb.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+
+    PATH: str
+    DATA_DIR: str
+    USE_AUTH_TOKEN: Optional[Union[bool, str]]
+
+    def setUp(self):
+        self.dataset = load_dataset(
+            self.PATH,
+            name="bigbio",
```
galtay, line 62: probably want to allow a little bit of flexibility here ... every once in a while we'll have one dataset that supports multiple tasks and so might have  `bigbio-translation` and `bigbio-kb` in one dataset.

tests/test_kb.py
```diff
@@ -0,0 +1,191 @@
...
+
+                for example in self.dataset[split]:
+
+                    example_text = _get_example_text(example)
+
+                    for entity in example["passages"]:
+                        for idx, (start, end) in enumerate(entity["offsets"]):
+                            self.assertEqual(
+                                example_text[start:end], entity["text"][idx]
+                            )
```
galtay, line 85: can we change the variable name from `entity` to `passage` ?

tests/test_kb.py
```diff
@@ -0,0 +1,191 @@
...
+                            texts=entity["text"],
+                        ):
+
+                            entity_id = entity["id"]
+                            errors.append(
+                                f"Example:{example_id} - entity:{entity_id} " + msg
+                            )
+
+        if len(errors) > 0:
+            self.fail(msg="\n".join(errors) + OFFSET_ERROR_MSG)
```
galtay, line 118: I really like the idea of collecting some examples to show to the user .. very cool!

sg-wbi:
Merged with https://github.com/bigscience-workshop/biomedical/pull/121

PR #136, opened 2022-02-23
Skang example fix

sunnnymskang:
[DRAFT] The branch is to adapt the old examples to our most update template AND schema convention. 
I've completely  adapting ParamedDataset(@ruisi-su - would like to get your reviews on this) and in progress of doing so with Chemprot and CellFinder.

### DIFF
--- a/examples/cellfinder.py
+++ b/examples/cellfinder.py
@@ -14,26 +14,7 @@
 # limitations under the License.
 
 """
-This is a template on how to implement a dataset in the biomedical repo.
 
-A thorough walkthrough on how to implement a dataset can be found here:
-https://huggingface.co/docs/datasets/add_dataset.html
-
-This script corresponds to Step 4 in the Biomedical Hackathon guide.
-
-To start, copy this template file and save it as <your_dataset_name>.py in an appropriately named folder within datasets. Then, modify this file as necessary to implement your own method of extracting, and generating examples for your dataset. 
-
-There are 3 key elements to implementing a dataset:
-
-(1) `_info`: Create a skeletal structure that describes what is in the dataset and the nature of the features.
-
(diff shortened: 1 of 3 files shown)


examples/cellfinder.py
```diff
@@ -147,45 +102,60 @@ def _info(self):
...
-
-        Populate "SplitGenerator" with `name` and `gen_kwargs`. Note:
-
-        - `name` can be: datasets.Split.<TRAIN/TEST/VALIDATION> or a string
-        - all keys in `gen_kwargs` can be passed to `_generate_examples()`. If your dataset has multiple files, you can make a separate key for each file, as shown below:
-
-        """
-
         # No need for a download location; just specify file location
         data_dir = os.getcwd()
```
galtay, line 160: if this is a local dataset, we should get `data_dir` from the `load_dataset` kwarg. 
```suggestion
        data_dir = self.config.data_dir
```

examples/paramed.py
```diff
@@ -220,36 +154,40 @@ def _split_generators(
...
-        The arguments to this function come from `gen_kwargs` returned in `_split_generators()`
+    def _generate_examples(self,
+                           filepath,
+                           zh_file,
+                           en_file,
+                           split):
 
-        The goal of this function is to perform operations on any of the keys of `gen_kwargs` that allow you to extract and process the data.
+        logger.info("generating examples from = %s", filepath)
+        zh_file = open(zh_file, "r")
```
galtay, line 164: might want to use context managers here 

```
with open(zh_file, "r") as fp:
    zh_lines = fp.readlines()
with open(en_file, "r") as fp:
    en_lines = fp.readlines()
```

galtay (commented):
looks close! added a few comments

PR #160, opened 2022-02-26
Unittests

hakunanatasha:
Merged @leonweber + Sam's unit-tests.

Does the following

- given _SUPPORTED_TASKS, checks if the appropriate schema for the task has the right keys
- gets statistics, for each split, around how many examples of each key is present
- checks if name is correct
- checks if text matches offsets 
- tests if global ids are unique
- tests if referenced IDs are correct

You can run as such: `python -m tests.test_bigbio --path path_to_your_dataset.py`

I love the offsets + global id test but as it stands very few to any datasets will pass this; many will also have offset problems. We should discuss before adding them (I think these are important tests to have).

TODO:
- REVIEW THIS CODE
- update the readme to show an example of how to test
- add CI/CD for unit-tests


CONTRIBUTING.md
```diff
@@ -105,20 +105,33 @@ For the `_info_` function, you will need to define `features` for your
...
 Please read the [Task Schemas](#task_schemas.md) to understand how each key should behave.
 
 Populate the information in the dataset according to this schema; some fields may be empty.
 
+To enable quality control, please add the following line in your file before the class definition:
+```python
+_SUPPORTED_TASKS = ["your_task_name_here"]
+```
+For ease, please refer to the following keywords, given your NLP task:
+1. **NER, RE, Event Extraction, NED/ Coreference resolution**: `"kb"`
```
galtay, line 128: I suggest we break out the tasks here b/c some datasets will support one or more of these tasks so `_SUPPORTED_TASKS` would have things like `ner`, `re`, `ee`, `ned`, `coref`, ....

hakunanatasha: sounds fine to me; this was something I made a pragmatic judgement call to make it easier to check schemas but I agree this is way better

I'd like to merge @leonweber's test here too cause he had a nice strategy where you specify the dataset name whereas mine just searches in the folder; it would be good to integrate these two ideas as we can either do a quick search through all the datasets (For later) OR have the ability to check a specific dataset + pass any necessary flags with argparse

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+            print("=" * 10)
+
+            counter = defaultdict(int)
+            for example in split:
+                for feature_name, feature in schema.items():
+                    if example.get(feature_name, None) is not None:
+                        if isinstance(feature, datasets.Value) or isinstance(
+                            feature_name, dict
+                        ):
+                            if example[feature_name]:
```
galtay, line 168: is the `isinstance(feature_name, dict)` correct above? i'm not sure a dict can be used as a key in `example`.

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+                    [
+                        key in example
+                        for key in schema.keys()
+                        if key not in opt_keys
+                    ]
+                )
+                self.assertTrue(
+                    mandatory_keys,
+                    "id/document_id/passages missing from bigbio view",
+                )
```
galtay, line 399: I sometimes handle this kind of thing with set subtraction ... if you like this formulation feel free to use it, 
```
missing_keys = set(needed_keys) - set(example.keys())
self.assertTrue(
    len(missing_keys) == 0,
   f"{missing_keys} are missing from bigbio view",
)
```

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+
+                elif task == "events":
+                    subkeys += ["events"]
+
+                    for key in subkeys:
+                        for attrs in schema[key]:
+
+                            self.assertTrue(
+                                self._check_subkey(example[key][0], attrs)
+                            )
```
galtay, line 440: maybe construct subkeys in the if/elif conditionals but pull these for loops out and perform afterwards?

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+    "question-answering": "qa",
+    "entailment": "entailment",
+    "paraphrasing": "text_to_text",
+    "summarization": "text_to_text",
+    "translation": "text_to_text",
+    "text_to_text": "text_to_text",
+    "classification": "text",
+    "text": "text",
+    "pairs": "pairs",
+}
```
galtay, line 51: maybe call this `_TASK_TO_SCHEMA` ? might not make sense with some of the keys being schemas but had to go back and check this definition a few times while reviewing the code below

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+
+_SCHEMA_MAP = {
+    "kb": kb_features,
+    "qa": qa_features,
+    "entailment": entailment_features,
+    "text_to_text": text2text_features,
+    "text": text_features,
+    "pairs": pairs_features,
+}
```
galtay, line 60: following the example above, would it make sense to call this one `_SCHEMA_TO_FEATURES` ?

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+
+        self.assertTrue(
+            datascript.exists(),
+            "Script not named " + self.NAME + ".py",
+        )
+
+    def test_schema(self, task: str):
+        """Search supported tasks within a dataset and verify big-bio schema"""
+
+        mapped_task = self._test_task_exists(task)
```
galtay, line 366: `mapped_schema = self._test_task_exists(task)` ?

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+
+        if mapped_task == "kb":
+            log.info("NER/NED/RE/Event Extraction/Coref Task")
+
+            schema = kb_features
```
galtay, line 371: `features = kb_features` ?

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+            self._check_keys(_SCHEMA_MAP[mapped_task], mapped_task)
+
+        elif mapped_task == "pairs":
+            log.info("Pair Labels Task")
+            self._check_keys(_SCHEMA_MAP[mapped_task], mapped_task)
+
+        else:
+            raise ValueError(
+                task
+                + " not recognized; only 'kb', 'qa', 'entailment', 'text_to_text', 'text', 'pairs'-type tasks are allowed"
```
galtay, line 480: f"{task} not recognized, only {set(_TASK_MAP.values())}-type tasks are allowed"

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+                        self._assert_ids_globally_unique(elem, ids_seen)
+                else:
+                    if k == "id":
+                        if not ignore_assertion:
+                            self.assertNotIn(v, ids_seen)
+                        ids_seen.add(v)
+
+        elif isinstance(collection, list):
+            for elem in collection:
+                self._assert_ids_globally_unique(elem, ids_seen)
```
galtay, line 542: yea, maybe we can make this a warning? ... on the fence ... sometimes it will be a a dataset error, but often it will be an implementation error too ... definitely we should produce a warning for participants .. maybe we can sort the actual problem out during final pass Q/A.

galtay (commented):
Lots of comments, please don't see them as hard requests, just some possible alternatives.

tests/test_bigbio.py
```diff
@@ -0,0 +1,660 @@
...
+import argparse
+from collections import defaultdict
+from difflib import ndiff
+from pathlib import Path
+from pprint import pformat
+from typing import Iterator, Optional, Union, Iterable, List
+
+sys.path.append(str(Path(__file__).parent.parent))
+
+import logging as log
```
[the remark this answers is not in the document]
galtay: PR is #184

tests/test_bigbio.py
hakunanatasha: @leonweber and sam wrote this function; i think the point is iterable not necessarily dict)

hakunanatasha:
@leonweber @galtay @sunnnymskang  please remove `test_kb_dataset` and `test_kb_dataset2` before merging if you merge; these are part of `test_bigbio`

Issue #148, opened 2022-02-25
Create dataset loader for BioNLP-ST 2013 PC


leonweber:
- **Name:** BioNLP-ST 2013 PC
- **Description:** Event extraction data set for pathway curation
- **Task:** Event extraction
- **Paper:** https://aclanthology.org/W13-2009.pdf
- **Data:** [https://github.com/openbiocorpora/bionlp-st-2013-pc]()
- **License:** License Terms for BioNLP Shared Task 2013 Data

barthfab:
#self-assign

PR #164, opened 2022-02-28
Add utils.py

leonweber:
Add utilities file that contains two functions for parsing brat files.

The brat parsing utilities were tested with MLLE (#77) and BioNLP-2013_PC (#148). 
Originally, it was planned to also include BioC parsing utilities in the utils, but @sg-wbi found that the BioC schema is only fixed for passages and annotations. All other BigBio-KB fields like normalisation, corefs or relations would require custom code. So we decided to skip the inclusion of BioC into the utils.

### DIFF
--- a/examples/utils.py
+++ b/examples/utils.py
@@ -1,0 +1,312 @@
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, Iterable
+
+
+def parse_brat_file(txt_file: Path) -> Dict:
+    """
+    Parse a brat file into the schema defined below.
+    `txt_file` should be the path to the brat '.txt' file you want to parse, e.g. 'data/1234.txt'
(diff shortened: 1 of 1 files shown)


examples/utils.py
```diff
@@ -0,0 +1,312 @@
...
+                                "string"
+                            ),  # Human readable description/name of the entity, e.g. "Barack Obama"
+                        }
+                    ],
+                },
+            )
+    """
+
+    example = {}
+    example["document_id"] = txt_file.name.removesuffix(".txt")
```
galtay, line 91: do brat files always have `txt` extensions? probably don't need to fix in this PR, just wondering if its a standard.

galtay (approved):
Looks good to me. I haven't tested it at all, but I'm OK merging

PR #165, opened 2022-02-28
Clarify id definition

galtay:
Describe id as unique within each dataset split.

see https://huggingface.slack.com/archives/C020T0W9PBK/p1646045120458169


jason-fries:
LGTM!

PR #167, opened 2022-03-01
Requirements fix

### LINKED ISSUE #132: Can't install requirements due to duplicate declaration of huggingface_hub in requirements.txt
## Describe the bug

I cannot install the requirements via pip due to a duplicate declaration of huggingface_hub in requirements.txt: `
huggingface_hub>=0.1.0,<1.0.0` and `huggingface_hub>=0.4.0`

## Steps to reproduce the bug
```bash
pip3 install -r requirements.txt
```

## Expected results

Installation of requirements.

## Actual results

```bash
ERROR: Double requirement given: huggingface_hub>=0.4.0 (from -r requirements.txt (line 11)) (already in huggingface_hub<1.0.0,>=0.1.0 (from -r requirements.txt (line 7)), name='huggingface-hub')
```

## Environment info

- `datasets` version:
- Platform: win64
- Python version: 3.7.9
- PyArrow version:

nomisto:
pip<=20.3 allows only one top level specification of a requirement. Requirements however specifies `huggingface_hub` two times. fixes #132

Futher info: https://github.com/pypa/pip/issues/2367, https://github.com/pypa/pip/issues/988

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ pandas==1.3.3
 numpy==1.21.2
 pybrat @ git+https://github.com/Yevgnen/pybrat
 datasets==1.18.3
-huggingface_hub>=0.4.0,<1.0.0
+huggingface_hub>=0.1.0,<1.0.0
 black~=22.0
 flake8>=3.8.3
 isort>=5.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,7 @@ pandas==1.3.3
 numpy==1.21.2
 pybrat @ git+https://github.com/Yevgnen/pybrat
 datasets==1.18.3
-huggingface_hub>=0.1.0,<1.0.0
+huggingface_hub>=0.4.0,<1.0.0
 black~=22.0
 flake8>=3.8.3
 isort>=5.0.0
-huggingface_hub>=0.4.0


galtay (changes requested):
both `datasets` and `transformers` use `huggingface_hub>=0.1.0,<1.0.0`

* https://github.com/huggingface/transformers/blob/master/setup.py#L111
* https://github.com/huggingface/datasets/blob/master/setup.py#L99

I'm not sure if those requirements are stale, but perhaps we should follow that example for now? thanks for finding the double requirement!

nomisto:
Sure, sounds good! I've changed the huggingface_hub version specifier to that of datasets. It would however be good to know what the reason was for `huggingface_hub>=0.4.0` in the first place in https://github.com/bigscience-workshop/biomedical/commit/9c8d2794b1828e8f08f11e4dc4fb3cd76ec4884d, not that we are introducing some bug again.

galtay-tempus:
@hakunanatasha thoughts here?

galtay-tempus (approved):
Going to merge the versioning that is consistent with `datasets` and `transformers` for now, but we can change to `>=0.4.0` if we need to later.

galtay-tempus (approved):

PR #173, opened 2022-03-02
pin pybrat instead of install from source?

galtay:
not sure if we were installing from source for a specific reason before, but maybe we should pin to a PyPI version?
(github editor added some whitesapce to last line)


sunnnymskang:
Have you tested building environments with the new requirements and did it work to load the datasets?

galtay-tempus:
I have not and have not developed any dataset loaders that used this package ... just figured its safer b/c the github code can change at any time ... was mostly wondering if whoever put the requirement in initially had thoughts on this?

galtay-tempus:
ps, it could totally be that the source is more up to date than the latest pinned version in PyPI

PR #174, opened 2022-03-03
Clean up for 3 examples data loaders

jason-fries:
- Added `BigBioBuilderConfig`
- Made VERSION constants consistent across examples
- Made config names more standardized
- Added copyright notice back to hallmarks_of_cancer
- Updated task_schemas to include standard abbreviations for _SUPPORTED_TASKS
- Tracking a global (across split) is slightly more complicated -- unique across split is fine.


examples/hallmarks_of_cancer.py
```diff
@@ -1,175 +1,189 @@
...
-                                'label': label_body
+                    "id": uid,
+                    "document_id": filenname.split(".")[0],
+                    "text": text_body,
+                    "label": label_body,
                 }
 
-                uid+=1
-
+                uid += 1
```
galtay-tempus, line 188: I think we'll want to update the `task_schemas.md` text around `id` fields being unique within each split (around line 45 in "Schema Notes")

examples/bc5cdr.py
```diff
@@ -68,32 +69,46 @@
...
-        datasets.BuilderConfig(
-            name="bigbio",
-            version="1.0.0",
-            description="BigScience biomedical hackathon schema",
+        BigBioConfig(
+            name="bc5cdr_bigbio",
+            version=_BIOBIO_VERSION,
+            description="BC5CDR simplified BigBio schema",
+            schema="bigbio_kb",
+            task_id="bc5cdr",
```
galtay-tempus, line 105: my impression when reading this config variable is that it would be something like "QA" or "NER" ... should we rename this to `dataset_name` maybe? if the goal is to capture things like the year in a BioASQ task then we could take it even further and have a BioASQ config that inherits from BigBioConfig and has a default value for a "years" field. that would allow it to be treated the same as all BigBio Configs but also give users the power to control which years are used.

PR #181, opened 2022-03-11
one bigbio config

galtay:
Example of using a custom BigBioConfig with attributes like, 
* schema  (source|kb|entailment|...)
* task (qa|translation|coref|...)
* subset e.g. 10b for bioasq

In this approach there is only one `name` argument that is ever used and its `bigbio` (i.e. the name of the single config) 
```
    BUILDER_CONFIGS = [
        BigBioConfig(
            name="bigbio",
            version=SOURCE_VERSION,
            description="Bigbio Config",
        )
    ]
```

All other branching logic in the dataset loader script can be handled by accessing the `schema`, `task`, or `subset` args. 
```
ds_orig = load_dataset("n2c2_2011_coref.py", name="bigbio", schema="source", data_dir="/path/to/n2c2_2011_coref")
ds_bb = load_dataset("n2c2_2011_coref.py", name="bigbio", schema="kb", data_dir="/path/to/n2c2_2011_coref")
```

### DIFF
--- a/examples/n2c2_2011_coref.py
+++ b/examples/n2c2_2011_coref.py
@@ -403,7 +403,7 @@ class N2C22011CorefDataset(datasets.GeneratorBasedBuilder):
                 }
             )
 
-        elif self.config.schema == "bigbio-kb":
+        elif self.config.schema == "kb":
             features = Features(
                 {
                     "id": Value("string"),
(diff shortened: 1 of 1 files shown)


galtay:
closing in favor of the config jason proposed.

PR #182, opened 2022-03-13
typing fixes, dedupe entities, text from offsets

galtay:
great unit tests! :) 

I was able to run all 3 on the n2c2_2011_coref dataset. 
It led me to discover a duplicate entity in the dataset and remove it. 
I also fixed up some entity text choices. Initially I used the text that was provided in the dataset for the entity. This text was lowercased compared to the text in the passages. This caused a lot of text mismatch errors. I switched the text to be from the passages themselves and added some notes in the dataset _loader implementation. 
I also updated the type hints from things like `list[list[int]]` to `List[List[int]]`.
The former was throwing errors in python 3.8 

* python -m tests.test_bigbio --path biomeddatasets/n2c2_2011_coref/n2c2_2011_coref.py --data_dir ~/Data/bigbio/n2c2-2011-coref
* python -m tests.test_kb_dataset --path biomeddatasets/n2c2_2011_coref/n2c2_2011_coref.py --data_dir ~/Data/bigbio/n2c2-2011-coref
* python -m tests.test_kb_dataset2 --path biomeddatasets/n2c2_2011_coref/n2c2_2011_coref.py --data_dir ~/Data/bigbio/n2c2-2011-coref

### DIFF
--- a/biomeddatasets/n2c2_2011_coref/n2c2_2011_coref.py
+++ b/biomeddatasets/n2c2_2011_coref/n2c2_2011_coref.py
@@ -267,9 +267,10 @@ def _tokoff_from_line(text: str) -> List[Tuple[int, int]]:
     return tokoff
 
 
-def _form_entity_id(sample_id, start_line, start_token, end_line, end_token):
-    return "{}-entity-{}-{}-{}-{}".format(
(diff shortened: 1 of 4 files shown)


sunnnymskang:
Thank you for confirming that unit-tests worked for your dataset : )

PR #190, opened 2022-03-16
Fix mapping from task to keys: ner -> entities, re -> entities/relations

leonweber:
Fix mapping from task to keys: ner -> entities, re -> entities/relations

Previous mapping was ner -> entities/relations, re -> entities/relations


galtay (approved):
:+1:

PR #195, opened 2022-03-18
Convert assertions that could fail due to data quality to warnings

leonweber:
The following tests could fail due to errors in the source data and thus now only fire warnings instead of hard assertion fails:
* test_do_all_referenced_ids_exist
* test_entities_offsets


galtay (approved):
I ran `python -m tests.test_bigbio --path examples/bc5cdr.py --schema KB` . It now yields a lot of information! nice work

PR #201, opened 2022-03-20
BC5CDR Fixes

jason-fries:
Fixed BC5CDR unit test failures for:

- Discontiguous entity offsets
- Missing relation ids

BC5CDR defines relations at the document-level, so this PR materializes relations as explicit entity mentions when using the bigbio schema.


galtay-tempus (approved):
looks great! tests pass on my end too

galtay (approved):

PR #204, opened 2022-03-21
Make mlee pass tests

leonweber:
This makes MLEE pass the unit test by fixing the following:
* `examples/` is now a python module to allow import of utils.py
* Fix a regression in `test_bigbio.py` that made `test_do_all_referenced_ids_exist` auto fail for event extraction data
* Fix assumption in `test_bigbio.py:test_schema` that the first example has annotations for all included keys
* Fix list of supported tasks for MLEE


sunnnymskang:
Confirmed that this won't change the result of the unit test for other datasets

PR #259, opened 2022-03-22
Fixes for BIOSSES

jason-fries:
Uploaded a converted version of the original Word doc files to the biomedical community hub.
Various template fixes


sunnnymskang:
Confirmed that it runs and passes unit tests

PR #288, opened 2022-03-30
Clarifications according to beta feedback 2

leonweber:
Improves a few things in the documentation according to the second round of beta feedback from @mariosaenger. I hope that the commit messages are self-explanatory

### DIFF
--- a/templates/template.py
+++ b/templates/template.py
@@ -218,6 +218,7 @@ class NewDataset(datasets.GeneratorBasedBuilder):
             data_dir = self.config.data_dir
 
         # Not all datasets have predefined canonical train/val/test splits. If your dataset does not have any splits, you can omit any missing splits.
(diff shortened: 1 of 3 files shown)


galtay (approved):

templates/template.py
```diff
@@ -218,6 +218,7 @@ def _split_generators(self, dl_manager) -> List[datasets.SplitGenerator]:
             data_dir = self.config.data_dir
 
         # Not all datasets have predefined canonical train/val/test splits. If your dataset does not have any splits, you can omit any missing splits.
+        # If your dataset has no predefined splits, use datasets.Split.TRAIN for all of the data.
```
galtay, line 221: This seems a bit confusing when coupled with the previous line. I would remove "If your dataset does not have any splits, you can omit any missing splits."

galtay (approved):
I'll just merge this and make the change I requested in a new PR b/c this is 95% good stuff.

PR #300, opened 2022-04-01
update FAQs

galtay:
* add some tips for finding licenses 
* clean up public dataset / private dataset instructions
* replace ref to cellfinder with n2c2_2011 and bioasq


debajyotidatta:
Great! Just merging this PR now since it contains important info.

PR #302, opened 2022-04-01
Add annotation file suffix parameter for brat parsing

mariosaenger:
Hi,

this PR makes it possible to specify more precisely which annotation files should be read while parsing corpora in the Brat annotation format. This is necessary, for example, if the corpus contains the annotations of two or more annotators that need to be read separately.

Best,
Mario

### DIFF
--- a/utils/parsing.py
+++ b/utils/parsing.py
@@ -1,6 +1,6 @@
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, Iterable
+from typing import Dict, Iterable, List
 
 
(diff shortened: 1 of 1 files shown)


galtay:
@leonweber can you take a look at this when you get a chance please.

galtay:
@leonweber can you take a look at this when you get a chance please.

leonweber (approved):
LGTM. Tests for MLEE still pass and the changes are necessary for BRAT datasets with slight deviations in naming conventions. Thanks @mariosaenger !

PR #324, opened 2022-04-04
Closes #26

### LINKED ISSUE #26: Create dataset loader from MedMentions
From https://github.com/chanzuckerberg/MedMentions

nomisto:
closes #26 

- **Name:** MedMentions
- **Description:** MedMentions is a new manually annotated resource for the recognition of biomedical concepts. 
What distinguishes MedMentions from other annotated biomedical corpora is its size (over 4,000
abstracts and over 350,000 linked mentions), as well as the size of the concept ontology (over
3 million concepts from UMLS 2017) and its broad coverage of biomedical disciplines.
- **Paper:** https://arxiv.org/abs/1902.09476
- **Data:** https://github.com/chanzuckerberg/MedMentions

Dataset supports two subsets "medmentions_full" and "medmentions_st21pv" (which is a subset of full).

Dataset was created as a "Named Entity Disambiguation" dataset, it can however be used for Named Entity Recognition aswell.

**Question:** In this dataset entities can have multiple types. For example NE `copper` has the three types `Pharmacologic Substance`, `Biologically Active Substance` and `Element, Ion, or Isotope`. In the KB-schema however `type` is a single string. I now concatenated (as did the source data) all types with ",". This feels however a little bit hacky, what should I do?

### DIFF
--- a/biodatasets/medmentions/medmentions.py
+++ b/biodatasets/medmentions/medmentions.py
@@ -219,16 +219,15 @@ class MedMentionsDataset(datasets.GeneratorBasedBuilder):
                     entities_ = []
                     for entity in document["entities"]:
                         for type in entity["semantic_type_id"]:
-                            entities_.append({
-                                "id": next(uid),
-                                "type": type,
-                                "text": entity["text"],
-                                "offsets": entity["offsets"],
-                                "normalized": [{
-                                    "db_name": "UMLS", 
-                                    "db_id": entity["concept_id"]
-                                }]
-                            })
+                            entities_.append(
+                                {
+                                    "id": next(uid),
+                                    "type": type,
+                                    "text": entity["text"],
+                                    "offsets": entity["offsets"],
+                                    "normalized": [{"db_name": "UMLS", "db_id": entity["concept_id"]}],
+                                }
+                            )
                     document["entities"] = entities_
 
                     for passage in document["passages"]:
--- a/biodatasets/medmentions/medmentions.py
+++ b/biodatasets/medmentions/medmentions.py
@@ -215,11 +215,22 @@ class MedMentionsDataset(datasets.GeneratorBasedBuilder):
                 for document in self._generate_parsed_documents(corpus, pmids):
                     document["id"] = next(uid)
                     document["document_id"] = document.pop("pmid")
+
+                    entities_ = []
                     for entity in document["entities"]:
-                        entity["id"] = next(uid)
-                        entity["type"] = ",".join(entity.pop("semantic_type_id"))
-                        concept_id = entity.pop("concept_id")
-                        entity["normalized"] = [{"db_name": "UMLS", "db_id": concept_id}]
+                        for type in entity["semantic_type_id"]:
+                            entities_.append({
+                                "id": next(uid),
+                                "type": type,
+                                "text": entity["text"],
+                                "offsets": entity["offsets"],
+                                "normalized": [{
+                                    "db_name": "UMLS", 
+                                    "db_id": entity["concept_id"]
+                                }]
+                            })
+                    document["entities"] = entities_
+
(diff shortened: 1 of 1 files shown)


leonweber:
Thanks! We're discussing the question wrt multiple entity types internally and will  post the solution here as soon as we have decided on something.

sg-wbi:
This case is now on the **FAQ in Discord**. This is how it should be handled:

> In many of our schemas, we have a 1:1 mapping between a key and its label (i.e. in KB, entity and label). In some datasets, we've noticed that there are multiple labels assigned to a text entity. Generally speaking, if a big-bio key has multiple labels associated with it, please populate the list with multiple instances of (key, label) according to each label that correspond to it. 
> 
> So for instance if the dataset has an entity "copper" with the  types "Pharmacologic Substance" and "Biologically Active", please create one entity with type "Pharmacologic Substance" and an associated unique id and another entity with type "Biologically Active" with a different unique id. The rest of the inputs (text, offsets, and normalization) of both entities will be identical.

nomisto:
Thanks! I've implemented these changes now and the dataset should be good to go now.

sg-wbi:
Thank you @nomisto!

Running the tests now gives the following error:
```
Traceback (most recent call last):
  File "/home/ele/Desktop/projects/tm/bigbio/tests/test_bigbio.py", line 153, in setUp
    self.dataset_source = datasets.load_dataset(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/load.py", line 1675, in load_dataset
    builder_instance = load_dataset_builder(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/load.py", line 1533, in load_dataset_builder
    builder_instance: DatasetBuilder = builder_cls(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 1020, in __init__
    super().__init__(*args, **kwargs)
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 258, in __init__
    self.config, self.config_id = self._create_builder_config(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 349, in _create_builder_config
    raise ValueError(f"BuilderConfig {name} not found. Available: {list(self.builder_configs.keys())}")
ValueError: BuilderConfig medmentions_source not found. Available: ['medmentions_full_source', 'medmentions_full_bigbio_kb', 'medmentions_st21pv_source', 'medmentions_st21pv_bigbio_kb']
```
Could you please take a look?

nomisto:
Hi @sg-wbi, the dataset has two subset ids `medmentions_full` and `medmentions_st21pv` (similar to bioasq_9b, bioasq_10b, ...). I think you have to run the tests with either, both these tests work:

```bash
python -m tests.test_bigbio biodatasets/medmentions/medmentions.py --subset_id medmentions_full
python -m tests.test_bigbio biodatasets/medmentions/medmentions.py --subset_id medmentions_st21pv
```

Help from the test function regarding the argument `subset_id`:
> by default, subset_id will be generated from path (e.g. if path=examples/n2c2_2011.py then subset_id=n2c2_2011). the config name is then constructed as config_name=<subset_id>_bigbio_<schema>. use this to explicitly set the subset_id for the config name you want to test (e.g. bioasq9b).

However, I don't know if it should work out of the box (without specifying --subset_id specifically) and this has to do something with my configuration? @leonweber @galtay

sg-wbi:
Pfff... sorry my bad. But let's hear their thoughts :+1:

hakunanatasha:
Hi @nomisto - in most cases out of the box exists since there is only one task and one main split; subset-id was intended to be there for explicitly this reason.

hakunanatasha (approved):

PR #328, opened 2022-04-04
Add dataloader for PDR corpus

mariosaenger:
Hi,

this PDR implements the plant-disease relation corpus (PDR). The corpus contains annotations from two different annotators which were implemented by different source views / schemes. The merged version is build by only keeping annotations on which both annotators agree. 

Best,
Mario

### DIFF
--- a/biodatasets/pdr/pdr.py
+++ b/biodatasets/pdr/pdr.py
@@ -1,0 +1,421 @@
+# coding=utf-8
+# Copyright 2022 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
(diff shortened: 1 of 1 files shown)


hakunanatasha:
Hi @mariosaenger thank you very much for your submission! I'll check this out ASAP.

hakunanatasha (dismissed):
Hi @mariosaenger 

Fantastic job. This is ready to go, but would you please mention somewhere in the script that the big-bio schema specifically corresponds to the merged annotations? This is in the PR but it's good have a record in the script.

hakunanatasha:
@mariosaenger I actually went ahead and added this one line - great job. I'm merging it.

PR #319, opened 2022-04-03
Create a dataset loader for DIANN IberEval 2018

danilexn:
Closes #63 

### Checkbox

- [x] Confirm that this PR is linked to the dataset issue.
- [x] Create the dataloader script `biodatasets/my_dataset/my_dataset.py` (please use only lowercase and underscore for dataset naming).
- [x] Provide values for the `_CITATION`, `_DATASETNAME`, `_DESCRIPTION`, `_HOMEPAGE`, `_LICENSE`, `_URLs`, `_SUPPORTED_TASKS`, `_SOURCE_VERSION`, and `_BIGBIO_VERSION` variables.
- [x] Implement `_info()`, `_split_generators()` and `_generate_examples()` in dataloader script.
- [x] Make sure that the `BUILDER_CONFIGS` class attribute is a list with at least one `BigBioConfig` for the source schema and one for a bigbio schema.
- [x] Confirm dataloader script works with `datasets.load_dataset` function.
- [x] Confirm that your dataloader script passes the test suite run with `python -m tests.test_bigbio biodatasets/my_dataset/my_dataset.py`.

PR #331, opened 2022-04-05
Tests: iterate over specified --schema in setUp

danilexn:
In #330, some changes were introduced to fix the functionality of `--schema`. However, running that version in #319 still returns the same errors specified in that discussion. 

This can be fixed by replacing `self._MAPPED_SCHEMAS` by `self.schemas_to_check` during BigBio dataset loading 

https://github.com/bigscience-workshop/biomedical/blob/02f02176a8773d7a13d81d443ba7101219669e2d/tests/test_bigbio.py#L161

### DIFF
--- a/tests/test_bigbio.py
+++ b/tests/test_bigbio.py
@@ -158,7 +158,7 @@ class TestDataLoader(unittest.TestCase):
         )
 
         self.datasets_bigbio = {}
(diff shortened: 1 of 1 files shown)


galtay (approved):
Thanks for catching this! I had this change locally but must have not pushed it.

PR #332, opened 2022-04-05
Closes #217

nomisto:
Closes #217

Task of dataset is Relation Extraction, however it was part of a challenge where the correct answers to the test set were never released. Since `data["test"]["relations"]` is thus empty for the testset the test fails with:

```bash
ERROR: runTest (__main__.TestDataLoader)
Run all tests that check:
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\ottsi\biomedical\tests\test_bigbio.py", line 116, in runTest
    self.test_schema(schema)
  File "C:\Users\ottsi\biomedical\tests\test_bigbio.py", line 518, in test_schema
    self.assertTrue(self._check_subkey(example[key][0], attrs))
IndexError: list index out of range

----------------------------------------------------------------------
Ran 1 test in 1.195s

FAILED (errors=1)
```

Edit: On second thought I think this might be a bug in the test-script, since it is assumed that a required subkey has to have elements in it. F.e. also in NER there could be documents without any entities at all in them.

PR #336, opened 2022-04-05
Fix existence check of required keys and make all tests subtests

leonweber:
This should fix the issue with the existence check of required keys (such as entities for NER datasets) mentioned in #332 

Changes are the following:
* Check that each split contains at least one instance of all required keys, as opposed to only the first example
* Fix bug that led to required keys only containing the keys for the first task (elifs were skipped => replaced with ifs)
* Make all tests subtests, so that tests will still run through after a failure. This is helpful for datasets where specific tests are expected to fail, e.g. for #332 where the test dataset will not contain annotations

@galtay @hakunanatasha @sg-wbi can you please have a look at this?

### DIFF
--- a/tests/test_bigbio.py
+++ b/tests/test_bigbio.py
@@ -7,7 +7,7 @@ import sys
 import unittest
 from collections import defaultdict
 from pathlib import Path
-from typing import Iterable, Iterator, List, Optional, Union
+from typing import Iterable, Iterator, List, Optional, Union, Dict
 
 from aiohttp.hdrs import TE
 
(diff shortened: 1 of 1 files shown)


tests/test_bigbio.py
```diff
@@ -7,7 +7,7 @@
 import unittest
 from collections import defaultdict
 from pathlib import Path
-from typing import Iterable, Iterator, List, Optional, Union
+from typing import Iterable, Iterator, List, Optional, Union, Dict
 
 from aiohttp.hdrs import TE
```
hakunanatasha, line 12: i just noticed we dont have this in the req file; just added it

hakunanatasha: or datasets lol 🤡

hakunanatasha (approved):
this looks fine to me; the only caveat is I don't see a case if features that are not supposed to be in the schema exist would be penalized.

I also added 2 changes to the requirements as it turns out I couldn't from-scratch use the unit-tests without them 🤡

tests/test_bigbio.py
```diff
@@ -100,22 +100,35 @@ def runTest(self):
         (10) test_coref_ids: Check if text matches offsets in coreferences
 
         """  # noqa
+        self.setUp()
```
galtay, line 103: this will run `setUp` twice right? https://docs.python.org/3/library/unittest.html#organizing-tests

leonweber: Yes, nice catch!

leonweber: Should be fixed by fcba9d8

galtay (dismissed):

hakunanatasha (approved):
@leonweber minor note

I would recommend moving the task-mapping (https://github.com/bigscience-workshop/biomedical/blob/fcba9d8da921c874769592b525358feb581c78e7/tests/test_bigbio.py#L492
) further north, prior to the declaration of the task (like `_TASK_TO_SCHEMA` https://github.com/bigscience-workshop/biomedical/blob/fcba9d8da921c874769592b525358feb581c78e7/tests/test_bigbio.py#L27)

I envision if we revisit this, we can potentially beef up this to consider other task-to-feature relationships. Currently now only KB is the unique situation, but I can see this populated in the future.

The var `_SCHEMA_TO_FEAUTURES` is misspelled; it should be `_SCHEMA_TO_FEATURES`. Given this is a very critical variable, I've opted to ignore this; if you feel comfortable replacing all instances correctly, i'd change it 🤡 
https://github.com/bigscience-workshop/biomedical/blob/fcba9d8da921c874769592b525358feb581c78e7/tests/test_bigbio.py#L45

Neither are breaking suggestions - carry on; this is approved.

Issue #140, opened 2022-02-24
Create a dataset loader for TwADR-L


sg-wbi:
- Task: NER, NEN
- License: Creative Commons Attribution 4.0 International
- Format: BioC
- Language: English
- Citation:
@inproceedings{limsopatham2016normalising,
  title={Normalising medical concepts in social media texts by learning semantic representation},
  author={Limsopatham, Nut and Collier, Nigel},
  booktitle={Proceedings of the 54th annual meeting of the association for computational linguistics (volume 1: long papers)},
  pages={1014--1023},
  year={2016}
}
- Source: https://zenodo.org/record/55013#.YgpAHVvMJhE

j-chim:
I'm working on #99 (same paper) and would be happy to work on this too. 
Is this open even though it's not marked as part of the hackathon?

sg-wbi:
Hi, thank you for contributing! Yes it is open and thanks for doble=checking with the official list. Don't know why it's not in there.

galtay:
@hakunanatasha can you add this to the project board?

PR #340, opened 2022-04-05
Closes #99

### LINKED ISSUE #99: Create a dataset loader for AskAPatient
- Task: NER, NEN
- License: Creative Commons Attribution 4.0 International
- Format: BioC
- Language: English
- Citation:
@inproceedings{limsopatham2016normalising,
  title={Normalising medical concepts in social media texts by learning semantic representation},
  author={Limsopatham, Nut and Collier, Nigel},
  booktitle={Proceedings of the 54th annual meeting of the association for computational linguistics (volume 1: long papers)},
  pages={1014--1023},
  year={2016}
}
- Source: https://zenodo.org/record/55013#.YgpAHVvMJhE

j-chim:
Please name your PR after the issue it closes. You can use the following line: "Closes #ISSUE-NUMBER" where you replace the ISSUE-NUMBER with the one corresponding to your dataset.

If the following information is NOT present in the issue, please populate:

- **Name:** *name of the dataset*
- **Description:** *short description of the dataset (or link to social media or blog post)*
- **Paper:** *link to the dataset paper if available*
- **Data:** *link to the online home of the dataset*

### Checkbox

### DIFF
--- a/biodatasets/ask_a_patient/ask_a_patient.py
+++ b/biodatasets/ask_a_patient/ask_a_patient.py
@@ -117,7 +117,7 @@ class AskAPatient(datasets.GeneratorBasedBuilder):
                 fold_id = re.search("AskAPatient\.fold-(\d)\.", fold_filepath).group(1)
                 document_id = f"{split_name}_{fold_id}"
                 split = datasets.SplitGenerator(
-                    name=f"{split_name}-{fold_id}",
+                    name=f"{split_name}_{fold_id}",
                     gen_kwargs={
                         "filepath": fold_filepath,
                         "document_id": document_id
--- a/biodatasets/ask_a_patient/ask_a_patient.py
+++ b/biodatasets/ask_a_patient/ask_a_patient.py
@@ -1,0 +1,149 @@
+# coding=utf-8
+# Copyright 2020 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import re
+import datasets
+import glob
+from pathlib import Path
+from utils import schemas
+from utils.configs import BigBioConfig
+from utils.constants import Tasks
+
+
+_DATASETNAME = "ask_a_patient"
+
+_CITATION = """
+@inproceedings{limsopatham-collier-2016-normalising,
+    title = "Normalising Medical Concepts in Social Media Texts by Learning Semantic Representation",
+    author = "Limsopatham, Nut  and
+      Collier, Nigel",
+    booktitle = "Proceedings of the 54th Annual Meeting of the Association for Computational Linguistics (Volume 1: Long Papers)",
+    month = aug,
+    year = "2016",
+    address = "Berlin, Germany",
+    publisher = "Association for Computational Linguistics",
+    url = "https://aclanthology.org/P16-1096",
+    doi = "10.18653/v1/P16-1096",
+    pages = "1014--1023",
+}
+"""
+
+_DESCRIPTION = """
+The AskAPatient dataset contains medical concepts written on social media \
+mapped to how they are formally written in medical ontologies (SNOMED-CT and AMT).
+"""
+
+_HOMEPAGE = "https://zenodo.org/record/55013"
+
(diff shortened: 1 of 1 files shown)


j-chim:
Questions to be resolved:
1. This is a text normalization dataset with (CUI, social media text, medical concept) tuples. I've put it as a paraphrasing task for now but want to check that it's appropriate.
2. How to incorporate the CUI into the BigBio schema? Putting it as "id" would break in cases when there are multiple expressions per medical concept and alternatives like putting it as "document_id"/prepending it to the texts etc will get messy.

galtay:
Took a peek at one of the files in the dataset and didn't immediately see any difference between `medical_concept` and `social_media_text` (the two were the same string for the examples I looked at). Do you know how often this is the case? If its a lot, this might not be great as a paraphrase task. This might fit into the kb schema though. Is there a way to determine which system the `cui` come from?

j-chim:
> Took a peek at one of the files in the dataset and didn't immediately see any difference between `medical_concept` and `social_media_text` (the two were the same string for the examples I looked at). Do you know how often this is the case? If its a lot, this might not be great as a paraphrase task. This might fit into the kb schema though. Is there a way to determine which system the `cui` come from?

Just checked this and it seems that, ignoring case, across all splits this is very common (~67%). I agree that paraphrase wouldn't be appropriate here. 
According to the paper, the `cui` comes from "SNOMED-CT and AMT", although there's no explicit information about which ontology was used/if there are overlaps.

sg-wbi:
I'd like to chime in here. This dataset is commonly used as NED and I think the KB schema is better suited for this one. The great overlap between text and normalization name is accounted for in the paper when they say
"""
We expect this dataset to be less difficult than TwADR-S and
TwADR-L, as the nature of blog posts is less informal and ambiguous than Twitter messages.
"""
My suggestion would be:
```python
{
"text" : social_media_text
"offsets" : [0, len(social_media_text)]
"normalized" : [{"db_name" : "SNOMED-CT/AMT", "db_id" : cui, "cui_name" : medical_concept}]
""
}
```
And drop "cui_name" for in the bigbio schema since the concept name is retrievalble from the "db_id". It;s a pity they do not distinguish between SNOMED and AMT but this is up to the end user anyway.

j-chim:
Thanks @sg-wbi !

For now I've kept the source schema the same as I think it would be helpful for users to access the data in a format that is closest to the original dataset released by the authors. 

For the bigbio schema I've updated the dataloader to use kb. If this looks ok I'll follow the same format for #140.

sg-wbi:
> For now I've kept the source schema the same as I think it would be helpful for users to access the data in a format that is closest to the original dataset released by the authors.

Yeah sorry, I am biased towards the bigbio scheme. "source" should be whatever feels more natural for the original dataset.

> For the bigbio schema I've updated the dataloader to use kb. If this looks ok I'll follow the same format for https://github.com/bigscience-workshop/biomedical/issues/140.
Sure, looks good to me!

biodatasets/ask_a_patient/ask_a_patient.py
```diff
@@ -0,0 +1,168 @@
...
+                            }
+                        ],
+                        "entities": [ 
+                            {
+                                "id": f"{id}_entity",
+                                "type": text_type,
+                                "text": [social_media_text],
+                                "offsets" : [offset],
+                                "normalized" : [
+                                    {"db_name" : "SNOMED-CT/AMT", "db_id" : cui}
```
galtay, line 161: minor thing, but what do you think about using "|" instead of "/" here between SNOMET-CT and AMT ... I'm hoping this wont be an issue for a lot of datasets, but when we are trying to map to these databases the symbol for "or" might make a little more sense.

j-chim: use of "|" makes sense to me, have updated it

biodatasets/ask_a_patient/ask_a_patient.py
```diff
@@ -0,0 +1,168 @@
...
+                                "offsets" : [offset],
+                                "normalized" : [
+                                    {"db_name" : "SNOMED-CT|AMT", "db_id" : cui}
+                                    ]
+                            }
+                        ],
+                        "events": [],
+                        "coreferences": [],
+                        "relations": [],
+                    }
```
sg-wbi, line 168: Can you fix this please? Then IMO it should be good to go. Thank you!

sg-wbi: It says "no newline at the end of the file"

j-chim: Can you post the stack trace here / on Discord please @sg-wbi? I don't see this when I run on my machine

sg-wbi: It;s not a bug, it's a "cosmetic" thing. If I go here: https://github.com/bigscience-workshop/biomedical/pull/340/files it shows at bottom of the file

j-chim: Ah! Many thanks for clarifying, fixed now

sg-wbi:
Thank you for contributing!!!

PR #342, opened 2022-04-05
allow multilabel text classification

galtay:
Update `label` -> `labels` in the text classification schema and update the hallmarks of cancer dataset.


jason-fries (approved):
LGTM!!

PR #345, opened 2022-04-06
make bioasq subset_ids have underscores

galtay:
the newer dataset loader scripts have implemented subset_ids with underscores. this makes bioasq consistent with that. 
@jason-fries can you take a look when you get a chance please. 

* underscores in subset_ids
* add LICENSE
* raise ValueError in missing data_dir


jason-fries (approved):
Looks great!

PR #366, opened 2022-04-07
Suggestion : new bigbio schema for text segmentation tasks

giganttheo:
I am suggesting a schema to use for text segmentation tasks such as POS-tagging. No other schema works for this kind of task for the moment from my understanding.


jason-fries:
Hi @giganttheo,
Thanks for the suggestions! There are definitely potential use cases for breaking out text segmentation into its own task + schema. However for this first round of the hackathon, we're focusing on trying to implement everything in the current schema set. If you have a POS task, you can either
- Implement just the `source` schema view
- Use the `bigbio_kb` schema and treat POS tags as word-level entities, as was done with [PICO Extraction](https://github.com/bigscience-workshop/biomedical/blob/master/biodatasets/pico_extraction/pico_extraction.py) This is verbose but should work fine!

We can revisit your PR once we're in the cleanup stage of the hackathon!

giganttheo:
> 

Ok, thanks for the response

> Hi @giganttheo, Thanks for the suggestions! There are definitely potential use cases for breaking out text segmentation into its own task + schema. However for this first round of the hackathon, we're focusing on trying to implement everything in the current schema set. If you have a POS task, you can either
> 
> * Implement just the `source` schema view
> * Use the `bigbio_kb` schema and treat POS tags as word-level entities, as was done with [PICO Extraction](https://github.com/bigscience-workshop/biomedical/blob/master/biodatasets/pico_extraction/pico_extraction.py) This is verbose but should work fine!
> 
> We can revisit your PR once we're in the cleanup stage of the hackathon!

Ok! Thanks for the response, I will do this!

PR #333, opened 2022-04-05
Closes #148

barthfab:
Please name your PR after the issue it closes. You can use the following line: "Closes #ISSUE-NUMBER" where you replace the ISSUE-NUMBER with the one corresponding to your dataset.

If the following information is NOT present in the issue, please populate:

- Name: BioNLP-ST 2013 PC
- Description: Event extraction data set for pathway curation
- Task: Event extraction
- Paper: https://aclanthology.org/W13-2009.pdf
- Data: [https://github.com/openbiocorpora/bionlp-st-2013-pc](https://github.com/bigscience-workshop/biomedical/issues/148)
- License: License Terms for BioNLP Shared Task 2013 Data

### Checkbox

- [x] Confirm that this PR is linked to the dataset issue.
- [x] Create the dataloader script `biodatasets/my_dataset/my_dataset.py` (please use only lowercase and underscore for dataset naming).
- [x] Provide values for the `_CITATION`, `_DATASETNAME`, `_DESCRIPTION`, `_HOMEPAGE`, `_LICENSE`, `_URLs`, `_SUPPORTED_TASKS`, `_SOURCE_VERSION`, and `_BIGBIO_VERSION` variables.
- [x] Implement `_info()`, `_split_generators()` and `_generate_examples()` in dataloader script.
- [x] Make sure that the `BUILDER_CONFIGS` class attribute is a list with at least one `BigBioConfig` for the source schema and one for a bigbio schema.
- [x] Confirm dataloader script works with `datasets.load_dataset` function.
- [x] Confirm that your dataloader script passes the test suite run with `python -m tests.test_bigbio biodatasets/my_dataset/my_dataset.py`.

PR #398, opened 2022-04-09
unit-test bypass args

hakunanatasha:
@leonweber @galtay 

I used @barthfab 's scripts as a template, since they are a good candidate of when they pass unit-tests except for the testing split which is understandably empty in the key that is to-be predicted.

Adds the following features for the unit-tests:

(1) Allows you to bypass testing for a split found in data generators (via ignoring in the `test_schema` for-loop)
(2) Allows you to bypass testing of a particular key in a schema (via removing from non_empty_features)
(3) Allows you to *ignore a single key within a specified split*

The defaults for all of these are empty.

You can test them as follows (I recommend using [this PR](https://github.com/bigscience-workshop/biomedical/pull/333) to test):

```
# This will fail
python -m tests.test_bigbio biodatasets/bionlp_st_2013_pc/bionlp_st_2013_pc.py 

# These will work
# Omit "events" in test
python -m tests.test_bigbio biodatasets/bionlp_st_2013_pc/bionlp_st_2013_pc.py --bypass_split_and_key test,events

python -m tests.test_bigbio biodatasets/bionlp_st_2013_pc/bionlp_st_2013_pc.py --bypass_splits test

python -m tests.test_bigbio biodatasets/bionlp_st_2013_pc/bionlp_st_2013_pc.py --bypass_keys events

# This will fail again, because only the test split is the issue
python -m tests.test_bigbio biodatasets/bionlp_st_2013_pc/bionlp_st_2013_pc.py --bypass_splits train
```

I included the (split, key) combo, because there are cases where you want to ensure most of the schema passes and you know a single key (or a few keys) may be problematic. The others can be used generally with admin approval

### DIFF
--- a/tests/test_bigbio.py
+++ b/tests/test_bigbio.py
@@ -78,14 +78,24 @@ class TestDataLoader(unittest.TestCase):
     """
     Given a dataset script that has been implemented, check if it adheres to the `bigbio` schema.
 
-    The test
-    """
+    Params:
+
+    :param PATH: path to dataloading script
+    :param SCHEMA: BigBio schema used to check info representation
+    :param SUBSET_ID: The explicit view/config to test BigBio on
+    :param DATA_DIR: If local, the path to dataset files
+    :param USE_AUTH_TOKEN: If private, an authentication token for loading data
+    :param BYPASS_SPLIT: If non-empty, splits in data to ignore testing
+    :param BYPASS_KEY: If non-empty, keys in data to ignore testing
+    """  # noqa
 
     PATH: str
     SCHEMA: str
     SUBSET_ID: str
     DATA_DIR: Optional[str]
     USE_AUTH_TOKEN: Optional[Union[bool, str]]
+    BYPASS_SPLIT: List[str]
+    BYPASS_KEY: List[str]
 
     def runTest(self):
         """
@@ -106,6 +116,9 @@ class TestDataLoader(unittest.TestCase):
         (10) test_coref_ids: Check if text matches offsets in coreferences
 
         """  # noqa
+
+        if len(self.BYPASS_SPLIT) > 0:
+            logger.warning("Splits ignored")
 
         for schema in self.schemas_to_check:
             dataset_bigbio = self.datasets_bigbio[schema]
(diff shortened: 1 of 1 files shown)


tests/test_bigbio.py
```diff
@@ -536,18 +606,97 @@ def test_schema(self, schema: str):
...
-            self.assertEqual(split.info.features, features)
-            for non_empty_feature in non_empty_features:
-                if split_to_feature_counts[split_name][non_empty_feature] == 0:
-                    raise AssertionError(f"Required key '{non_empty_feature}' does not have any instances")
+        split_to_feature_counts = self.get_feature_statistics(
+            features=features, schema=schema
+        )
+
+        # Omit a particular key in a split
+        if len(self.BYPASS_SPLIT_AND_KEY) > 0:
```
galtay, line 614: this seems like a lot of extra code ... is it possible to frame this as an early return condition in the original loop? 
```
for split_name, split in self.datasets_bigbio[schema].items():
    if split_name in self.BYPASS_SPLITS:
        continue  ? 
    self.assertEqual(split.info.features, features)
    for non_empty_feature in non_empty_features:
        if split_to_feature_counts[split_name][non_empty_feature] == 0:
            raise AssertionError(f"Required key '{non_empty_feature}' does not have any instances")
```

hakunanatasha: `BYPASS_SPLIT_AND_KEY` is a specific condition

In the above implementation, you can either bypass a data split (i.e. ignore ALL testing for "train"), bypass a key (i.e. "ignore events"). This applies uniformly to all cases.

In certain cases (i.e. many of the bionlp datasets), the test set is the only dataset to have only 1 key missing (for example, the test set is missing "events" in an event extraction task). For this case, it's wasteful and possibly dangerous to waive testing on the ENTIRE split, when the input examples (like entities) may need to be checked. The logic splits because in certain cases you need a *specific* combination of (split name, key to ignore), hence the overhead here.

If you have a suggestion that can streamline it, please feel free. This isn't written particularly efficiently but it works on the cases I've tested to ensure both the new args, and old behavior is maintained.

galtay: I might be missing something (and if this implementation is totally off, we can ignore it) but I was thinking all 3 conditions could be incorporated with early returns .... 
```
for split_name, split in self.datasets_bigbio[schema].items():

    # skip entire split
    if split_name in self.BYPASS_SPLITS:
        logger.info(f"skipping {split_name}")
        continue  

    self.assertEqual(split.info.features, features)

    for non_empty_feature in non_empty_features:
        # skip specific split,key combo 
        split_key = [split_name, non_empty_feature]
        if split_key in self.BYPASS_SPLIT_KEY:
            logger.info(f"skipping {split_key}")

        # skip key in every split
        if non_empty_features in self.BYPASS_KEYS:
            logger.info(f"skipping {non_empty_feature}")   

        if split_to_feature_counts[split_name][non_empty_feature] == 0:
            raise AssertionError(f"Required key '{non_empty_feature}' does not have any instances")
```

galtay: ps, the above was just me trying to do some code re-factor ... if we want to just merge these updates as is and cleanup after the hackathon, I'm OK with that too!

hakunanatasha: this is good - i dont mind implementing this albeit AT earliest sometime tomorrow. If you can replace the logic feel free to pull the branch or edit directly on the file.

hakunanatasha: code-refactors are good; as mentioned earlier, I wrote this pretty inelegantly to just get it done. Always good to have a second pair of eyes to squash some of the boilerplate.

sg-wbi:
This PR #453 has multiple `subset_id`s. It has 2 schemas (TEXT and KB), however these are mutually exclusive, i.e. one `subset_id` has only TEXT and another has KB. When I run the test I get this error:
```bash
ValueError: BuilderConfig codiesp_p_bigbio_kb not found. Available: ['codiesp_d_source', 'codiesp_p_source', 'codiesp_x_source', 'codiesp_extra_mesh_source', 'codiesp_extra_cie_source', 'codiesp_d_bigbio_text', 'codiesp_p_bigbio_text', 'codiesp_x_bigbio_kb', 'codiesp_extra_mesh_bigbio_text', 'codiesp_extra_cie_bigbio_text']
```
but for `subset_id` "codiesp_p" there is not supposed to be a "codiesp_p_bigbio_kb". Am I doing something wrong or does this warrant a a `bypass_schema` flag?

sg-wbi:
For reference: https://github.com/bigscience-workshop/biomedical/pull/488#issuecomment-1110340963

hakunanatasha:
Superceded by #533

PR #401, opened 2022-04-10
Closes #218

### LINKED ISSUE #218: Create dataset loader for PUBHEALTH
## Adding a Dataset
- **Name:** PUBHEALTH
- **Description:** *None provided*
- **Task:** SENT_CLASS
- **Paper:** https://aclanthology.org/2020.emnlp-main.623/
- **Data:** https://github.com/neemakot/Health-Fact-Checking/blob/master/data/DATASHEET.md
- **License:** MIT License

sidkiblawi:
Closes #218 

- **Name:** PUBHEALTH
- **Description:**A dataset of 11,832 claims for fact- checking, which are related a range of health topics.
- **Paper:**  https://aclanthology.org/2020.emnlp-main.623/
- **Data:** [*link to the online home of the dataset*](https://github.com/neemakot/Health-Fact-Checking/blob/master/data/DATASHEET.md)

### Checkbox

- [x] If my dataset is local, I have provided an output of the unit-tests in the PR (please copy paste). This is OPTIONAL for public datasets, as we can test these without access to the data files.

### DIFF
--- a/biodatasets/pubhealth/pubhealth.py
+++ b/biodatasets/pubhealth/pubhealth.py
@@ -53,7 +53,7 @@ _LICENSE = "MIT License"
 
 _URLs = {_DATASETNAME: "https://drive.google.com/uc?export=download&id=1eTtRs5cUlBP5dXsx-FTAlmXuB6JQi2qj"}
 
-_SUPPORTED_TASKS = [Tasks.TEXT_CLASSIFICATION]
+_SUPPORTED_TASKS = [Tasks.TEXTUAL_ENTAILMENT]
 _SOURCE_VERSION = "1.0.0"
 _BIGBIO_VERSION = "1.0.0"
 
@@ -73,10 +73,10 @@ class PUBHEALTHDataset(datasets.GeneratorBasedBuilder):
             subset_id="pubhealth",
         ),
         BigBioConfig(
-            name="pubhealth_bigbio_text",
+            name="pubhealth_bigbio_te",
             version=BIGBIO_VERSION,
             description="PUBHEALTH BigBio schema",
-            schema="bigbio_text",
+            schema="bigbio_te",
             subset_id="pubhealth",
         ),
     ]
@@ -100,9 +100,9 @@ class PUBHEALTHDataset(datasets.GeneratorBasedBuilder):
                 }
             )
 
-        # Using in pairs schema
-        elif self.config.schema == "bigbio_text":
-            features = schemas.text_features
+        # Using in entailment schema
+        elif self.config.schema == "bigbio_te":
+            features = schemas.entailment_features
 
         return datasets.DatasetInfo(
             description=_DESCRIPTION,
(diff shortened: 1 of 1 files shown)


biodatasets/pubhealth/pubhealth.py
```diff
@@ -0,0 +1,190 @@
...
+
+        with open(filepath, encoding="utf-8") as csv_file:
+            csv_reader = csv.reader(
+                csv_file, quotechar='"', delimiter="\t", quoting=csv.QUOTE_ALL, skipinitialspace=True
+            )
+            next(csv_reader, None)  # remove column headers
+            for id_, row in enumerate(csv_reader):
+                if split == "train" or split == "validation":
+                    if len(row) != 9:
+                        continue
```
galtay, line 144: would it be possible to put in a comment about why this is needed? 
I did a little exploration and I did see one example that looked malformed (in train file and starts with `What's false: However, these guidelines affected` ... appears to be that the claim_id is missing)  ... maybe even record with line numbers in each file are being skipped?

galtay (changes requested):

biodatasets/pubhealth/pubhealth.py
sidkiblawi: Added a logger to log the malformed lines.

biodatasets/pubhealth/pubhealth.py
```diff
@@ -0,0 +1,183 @@
...
+                        "label": label,
+                        "subjects": subjects,
+                    }
+
+                elif self.config.schema == "bigbio_text":
+                    yield id_, {
+                        "id": id_,  # uid is an unique identifier for every record that starts from 0
+                        "document_id": claim_id,
+                        "text": main_text,
+                        "labels": [label],
```
galtay, line 182: I'm just browsing the paper associated with this task and I'm not sure it makes sense as a (main_text, label) task. It seems that the claim, explanation, and main_text need to work together somehow to produce the label. What do you think?

sidkiblawi: Good point. I think if you look at figure 1 in the paper the text classification phase (veracity) utilizes both the claim and the explanation. So maybe we can create the "text" field to capture both by inserting something like: 
f"CLAIM: {claim} EXPLANATION: {explanation}"

sidkiblawi: Does that sound good or do you have other ideas?

sidkiblawi: Yes, I can see it. I'll adjust the bigbio_schema and resubmit.

galtay: did you put some thought into this part?
```
if we cant figure out a mapping between the existing labels [true, false, unproven, mix] and [entailment, neutral, contradiction]
```

sidkiblawi: Apologies, I totally misunderstood what you meant in that. I don't think we can make this an entailment task and instead need to make it a pairs task, since the explanation is used to verify the claim. In the instances the explanation is used to explain the label for the hypothesis. So I don't think that actually fits entailment.

galtay: No problem, I think you could make a case for mapping `true` to `entailment`, `false` to `contradiction` and collapsing `unproven` and `mix` into `neutral` but I think the text pairs version retains more of the info in the dataset so I think i'm happy either way. If you want to implement as text pairs with 4 possible values for the label that works for me. ps consider using  https://huggingface.co/docs/datasets/v2.0.0/en/package_reference/main_classes#datasets.ClassLabel in the source schema to indicate the 4 possible values (an example from BigBio https://github.com/bigscience-workshop/biomedical/blob/beac33de5b4044d718e3719832053a4315c24da8/biodatasets/meddialog/meddialog.py#L136)

galtay (approved):

galtay:
:tada:

PR #404, opened 2022-04-10
Closes #153 

### LINKED ISSUE #153: Create dataset loader for Bio-SimVerb
- **Name:** Bio-SimVerb
- **Description:** Verb pairs with similarity scores
- **Task:** Semantic Similarity
- **Paper:** https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-018-2039-z
- **Data:** https://github.com/cambridgeltl/bio-simverb/tree/master/wvlib/word-similarities/bio-simverb
- **License:** ?
- **Motivation:** Evaluation

debajyotidatta:
Closes #153 

### Checkbox

- [X] Confirm that this PR is linked to the dataset issue.
- [X] Create the dataloader script `biodatasets/my_dataset/my_dataset.py` (please use only lowercase and underscore for dataset naming).
- [X] Provide values for the `_CITATION`, `_DATASETNAME`, `_DESCRIPTION`, `_HOMEPAGE`, `_LICENSE`, `_URLs`, `_SUPPORTED_TASKS`, `_SOURCE_VERSION`, and `_BIGBIO_VERSION` variables.
- [X] Implement `_info()`, `_split_generators()` and `_generate_examples()` in dataloader script.
- [X] Make sure that the `BUILDER_CONFIGS` class attribute is a list with at least one `BigBioConfig` for the source schema and one for a bigbio schema.
- [X] Confirm dataloader script works with `datasets.load_dataset` function.
- [ ] Confirm that your dataloader script passes the test suite run with `python -m tests.test_bigbio biodatasets/my_dataset/my_dataset.py`.
- [X] If my dataset is local, I have provided an output of the unit-tests in the PR (please copy paste). This is OPTIONAL for public datasets, as we can test these without access to the data files.

Note: Data download path is longer than maximum allowed formatter in black. Other than that all the other tests pass.

### DIFF
--- a/biodatasets/bio_sim_verb/bio_sim_verb.py
+++ b/biodatasets/bio_sim_verb/bio_sim_verb.py
@@ -1,0 +1,175 @@
+# coding=utf-8
+# Copyright 2022 The HuggingFace Datasets Authors and the current dataset script contributor.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+This repository contains the evaluation datasets for the paper Bio-SimVerb and
+Bio-SimLex: Wide-coverage Evaluation Sets of Word Similarity in Biomedicine by
+Billy Chiu, Sampo Pyysalo and Anna Korhonen.
+"""
+
+import csv
+from typing import Dict, List, Tuple
+
+import datasets
+
+from utils import schemas
+from utils.configs import BigBioConfig
+from utils.constants import Tasks
+
+# TODO: Add BibTeX citation
+_CITATION = """\
+@article{chiu2018bio,
+  title={Bio-SimVerb and Bio-SimLex: wide-coverage evaluation sets of word similarity in biomedicine},
+  author={Chiu, Billy and Pyysalo, Sampo and Vuli{\'c}, Ivan and Korhonen, Anna},
+  journal={BMC bioinformatics},
+  volume={19},
+  number={1},
+  pages={1--13},
+  year={2018},
+  publisher={BioMed Central}
+}
+"""
+
+_DATASETNAME = "bio_sim_verb"
+
+
+_DESCRIPTION = """
+This repository contains the evaluation datasets for the paper
+Bio-SimVerb and Bio-SimLex: Wide-coverage Evaluation Sets of Word
+Similarity in Biomedicine by Billy Chiu, Sampo Pyysalo and Anna Korhonen.
+"""
+
+_HOMEPAGE = "https://github.com/cambridgeltl/bio-simverb"
+
+_LICENSE = """Open Access This article is distributed under the terms of the
+Creative Commons Attribution 4.0 International License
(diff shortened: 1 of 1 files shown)


debajyotidatta:
@galtay : Hey can you take a look at this PR? The test fails because the data download link is longer than the allowed character limit of black.

biodatasets/bio_sim_verb/bio_sim_verb.py
```diff
@@ -0,0 +1,175 @@
...
+    DEFAULT_CONFIG_NAME = "bio_sim_verb_source"
+
+    def _info(self) -> datasets.DatasetInfo:
+
+        if self.config.schema == "source":
+            features = datasets.Features(
+                {
+                    "text_1": datasets.Value("string"),
+                    "text_2": datasets.Value("string"),
+                    "label": datasets.Value("string"),
```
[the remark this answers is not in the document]
debajyotidatta: Why though? Isn't it better to stick to [consistent schemas](https://github.com/bigscience-workshop/biomedical/blob/master/utils/schemas/pairs.py) ? Secondly, for a whole set of tasks, it is much easier to generalize. For e.g ['not_equivalent', 'equivalent'] or any sort of binary non-string labels? Downstream these can easily be mapped.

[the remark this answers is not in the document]
galtay: I can totally see both interpretations but I slightly prefer the following, 
* use float for the source schema b/c it is always a float and that is what someone just using this one dataset would probably expect 
* use string in the bigbio schema b/c that is the schema we chose for multi-task approaches where it is sometimes more convenient to treat everything as a string.

biodatasets/bio_sim_verb/bio_sim_verb.py
```diff
@@ -0,0 +1,175 @@
...
+                uid = 0
+                for id_, row in enumerate(csv_reader):
+                    uid += 1
+                    text_1, text_2, label = row
+                    yield id_, {
+                        "id": str(uid),
+                        "document_id": "NULL",
+                        "text_1": text_1,
+                        "text_2": text_2,
+                        "label": "0" if float(label) < 5 else "1",
```
galtay, line 174: haha, here I was actually thinking just the string representation of the float (will have to go back through my PR comments but I think I suggested that in a different `bio_sim_X` PR)

galtay: ok, yep, it was @tanmaylaud 's PR for `bio-simlex` #406 :) yea .... I think I'm still in the "use float for source schema" camp ... but @debajyotidatta what do you think, is this thresholding how things were handled in the prompt source work?

hakunanatasha: @debajyotidatta I'd prefer we don't threshold the dataset. The users of the dataloader will expect the dataset to preserve as much streamlined information as possible. There is no guarantee this transformation actually preserves a well-balanced binary label class, and subtleties like scores close to 5 (i.e if the score distribution is gaussian in the center of the score range) will be lost.

Let's please do float for the source schema and str for the bigbio schema (to adhere to our schema unit-checks).

hakunanatasha:
@galtay @debajyotidatta commenting for future documentation purposes:

BigBio text schema for `label` will require a str; the score here is a float in theory. The source schema uses float and the label of BB uses str.

galtay (approved):

galtay:
see also $406

nomisto:
Hi @debajyotidatta, really small thing: There are currently the two datasets `bio_sim_verb` and `bio_simlex` which are basically related (Same author, one contains verbs, other one nomens). However, my inner Monk is triggered that they have different naming schemas. Could you rename yours to `bio_simverb`? If I do it, it will lose all its history and I would be the only contributor to this file and I don't want that.

PR #412, opened 2022-04-10
a constant to represent null values

galtay:
allows for, 
```
from utils.constants import BigBioValues

BigBioValues.NULL
```


jason-fries (approved):
LGTM!

PR #424, opened 2022-04-11
Closes #97

### LINKED ISSUE #97: Create a dataset loader for Verspoor2013
- Task: NER, RE
- License: ?
- Format: brat
- Language: English
- Citation:
@article{verspoor2013annotating,
  title={Annotating the biomedical literature for the human variome},
  author={Verspoor, Karin and Jimeno Yepes, Antonio and Cavedon, Lawrence and McIntosh, Tara and Herten-Crabb, Asha and Thomas, Zo{\"e} and Plazzer, John-Paul},
  journal={Database},
  volume={2013},
  year={2013},
  publisher={Oxford Academic}
}
- Source: https://github.com/rockt/SETH/tree/master/resources/Verspoor2013

mcullan:
Please name your PR after the issue it closes. You can use the following line: "Closes #ISSUE-NUMBER" where you replace the ISSUE-NUMBER with the one corresponding to your dataset.

If the following information is NOT present in the issue, please populate:

- **Name:** *name of the dataset*
- **Description:** *short description of the dataset (or link to social media or blog post)*
- **Paper:** *link to the dataset paper if available*
- **Data:** *link to the online home of the dataset*

### Checkbox

- [x] If my dataset is local, I have provided an output of the unit-tests in the PR (please copy paste). This is OPTIONAL for public datasets, as we can test these without access to the data files.


mcullan:
I have a few notes and questions on this one:

1. Still need to finish the description, so please ignore that for now.
2. I'm currently calling everything Verspoor2013, verspoor_2013, etc, because that seems to be the best name for the dataset that I could find.
3. This can be automatically downloaded through the Github repo linked in the original issue. It's a bit funny to do so, so I wanted to check if that's alright or if I should make it local instead. It seems funny because requires downloading other datasets too (because you can only download a zipfile of the entire parent repo.
4. For automatic downloads to work, I needed to modify `utils.parsing.parse_brat_file` to accept separate directories for annotations and text files. I could avoid this by making it a local dataset, but would need to tell the user to move the text files and annotations into a single directory. I changed it in a way that shouldn't break existing code. It just now takes an optional argument for a directory where we can find annotation files.

sg-wbi:
> I'm currently calling everything Verspoor2013, verspoor_2013, etc, because that seems to be the best name for the dataset that I could find.

Sounds right to me!

> This can be automatically downloaded through the Github repo linked in the original issue. It's a bit funny to do so, so I wanted to check if that's alright or if I should make it local instead. It seems funny because requires downloading other datasets too (because you can only download a zipfile of the entire parent repo.

Yeah this is a bit unfortunate. The original dataset was published [here (see link in last line of the abstract)](https://pubmed.ncbi.nlm.nih.gov/23584833/), however the link it's broken and we have to resort to that github repo. Maybe there is a way to get only that sub-directory? But I wouldn't over-complicate things...

> For automatic downloads to work, I needed to modify utils.parsing.parse_brat_file to accept separate directories for annotations and text files. I could avoid this by making it a local dataset, but would need to tell the user to move the text files and annotations into a single directory. I changed it in a way that shouldn't break existing code. It just now takes an optional argument for a directory where we can find annotation files.

Ideally we should not modify `utils.parsing.parse_brat_file` unless strictly necessary (e.g. a bug). But still I would not make this a local dataset. Could you implement the moving of the files directly in the dataloader script? That to me would be the best solution.

biodatasets/verspoor_2013/verspoor_2013.py
```diff
@@ -0,0 +1,252 @@
...
+
+_DESCRIPTION = """\
+This dataset is designed for XXX NLP task.
+"""
+
+_HOMEPAGE = ""
+
+# The source repo (https://github.com/rockt/SETH/blob/master/LICENSE) is license
+# Apache 2.0 but does that match the dataset? Best info I have for now...
+_LICENSE = """\
```
sg-wbi, line 52: > he source repo (https://github.com/rockt/SETH/blob/master/LICENSE) is license
 Apache 2.0 but does that match the dataset?

Exactly: we do not know. I would put a `utils.constants.BigBioValues.NULL` here

sg-wbi (commented):
@mcullan Thank you for your contribution! The scripts looks good to me! Once we resolve the minor things you pointed out to it should be good to go

biodatasets/verspoor_2013/verspoor_2013.py
```diff
@@ -0,0 +1,252 @@
...
+            features=features,
+            # homepage=_HOMEPAGE,
+            license=_LICENSE,
+            citation=_CITATION,
+        )
+
+    def _split_generators(self, dl_manager) -> List[datasets.SplitGenerator]:
+        """Returns SplitGenerators."""
+
+        data_dir = Path(dl_manager.download_and_extract(_URLS[0]))
```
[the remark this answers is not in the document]
mcullan: Will do!

biodatasets/verspoor_2013/verspoor_2013.py
mcullan: Got it!

mcullan:
> Maybe there is a way to get only that sub-directory? But I wouldn't over-complicate things...
I believe there's an option that uses git to do a [sparse checkout](https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/), but that definitely sounds like over-complicating things to me. Since I'm adding some logic in `_split_generators` to move the annotations and data into the same folder, I could also delete the unused files as well.

sg-wbi:
> Since I'm adding some logic in _split_generators to move the annotations and data into the same folder, I could also delete the unused files as well.

It's not necessary :) the files are extremely small anyway.

mcullan:
@sg-wbi Whoops, I sort of accidentally clicked re-request review. Sorry about that!

sg-wbi:
Hey @mcullan I am sorry, I think I screwed up this PR by removing the `parsing.py` file. Could you please open a new one for the same dataset with only the `verspoor_2013.py` file? I tried to remove the unwanted `parsing.py` file myself to speed up things...

Issue #137, opened 2022-02-24
Create dataset loader for Mantra GSC


jason-fries:
## Adding a Dataset
- **Name:** *Mantra GSC*
- **Description:** *We selected text units from different parallel corpora (Medline abstract titles, drug labels, biomedical patent claims) in English, French, German, Spanish, and Dutch. Three annotators per language independently annotated the biomedical concepts, based on a subset of the Unified Medical Language System and covering a wide range of semantic groups.*
- **Task:** *NER, NED*
- **Paper:** [*A multilingual gold-standard corpus for biomedical concept recognition: the Mantra GSC*](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4986661/)
- **Data:** *[https://files.ifi.uzh.ch/cl/mantra/gsc/GSC-v1.1.zip](https://files.ifi.uzh.ch/cl/mantra/gsc/GSC-v1.1.zip)*
- **License:** *[CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)*
- **Motivation:** *This is a multilingual, parallel dataset (French, Spanish, Dutch, German, English)*

karthikrangasai:
Hello all,

The dataset has a brat format available, but the `.ann` files have lines starting with `#` which according to the format specification are Annotator Notes. ( [Format Specification](https://brat.nlplab.org/standoff.html) )

But in the examples of the dataset, the Tags seem to be in that very line that starts with `#`.

I am unsure of how to proceed in regards with this and am seeking some help.

Thank you.

![Screenshot from 2022-04-06 19-46-09](https://user-images.githubusercontent.com/39360170/162000295-0d552939-a327-4626-a3ff-3c673f3426d7.png)

hakunanatasha:
@leonweber @jason-fries @galtay @sg-wbi @sunnnymskang @debajyotidatta Pinged some other admins to answer this!

leonweber:
From the paper: The comments correspond to 'CUI, preferred term, semantic type, and semantic group (DISO is “Disorders”)'. These have been all manually verified. My suggestion would be to
* Add support for annotator notes to `utils.parsing.parse_brat`
* Implement 'CUI', 'preferred term', 'semantic type' and 'semantic group' all as separate fields in the `source` schema
* Implement 'semantic group' as entity type in the `bigbio_kb` schema

What do you think @karthikrangasai? Does that sound reasonable?

karthikrangasai:
Hello @leonweber ,

[...]

Issue #72, opened 2022-01-21
Create a dataset loader for CANTEMIST (CANcer TExt Mining Shared Task)


hakunanatasha:
From https://temu.bsc.es/cantemist/

hakunanatasha:
Hi @mapama247, can you let us know if you are still working on this so we can update our project board? Please just notify us the status **by Friday April 8**, no worries if you are not finished but intend to work on this. Please either ping me here at @hakunanatasha  or ping the discord admins (with @admins)

mapama247:
Hi @hakunanatasha! Yes, I have not finished yet but I plan to do it before the deadline ends.

PR #443, opened 2022-04-13
BRAT: Add support for annotator notes to brat parser

leonweber:
This adds support for annotator notes to the brat parser, because these notes contain annotations in some cases (#72, #137). Notes are only parsed when `parse_notes == True`. `parse_notes` is `False` by default to maintain backwards compatibility.


galtay (approved):
LGTM

PR #364, opened 2022-04-07
Closes #246

danilexn:
Closes #246 

- **Description:** PsyTAR dataset contains 891 drugs reviews posted by patients on "askapatient.com", about the effectiveness and adverse drug events associated with Zoloft, Lexapro, Cymbalta, and Effexor XR.

### Checkbox

- [x] Confirm that this PR is linked to the dataset issue.
- [x] Create the dataloader script `biodatasets/my_dataset/my_dataset.py` (please use only lowercase and underscore for dataset naming).
- [x] Provide values for the `_CITATION`, `_DATASETNAME`, `_DESCRIPTION`, `_HOMEPAGE`, `_LICENSE`, `_URLs`, `_SUPPORTED_TASKS`, `_SOURCE_VERSION`, and `_BIGBIO_VERSION` variables.
- [x] Implement `_info()`, `_split_generators()` and `_generate_examples()` in dataloader script.
- [x] Make sure that the `BUILDER_CONFIGS` class attribute is a list with at least one `BigBioConfig` for the source schema and one for a bigbio schema.
- [x] Confirm dataloader script works with `datasets.load_dataset` function.
- [x] Confirm that your dataloader script passes the test suite run with `python -m tests.test_bigbio biodatasets/my_dataset/my_dataset.py`.

PR #454, opened 2022-04-14
[WIP dont merge] require name in config

galtay:
Make `name` a required argument in `load_dataset` for our `BigBioConfig` 

see discussion here https://github.com/bigscience-workshop/biomedical/pull/364#issuecomment-1096845673


galtay:
we can totally defer this until later as explicitly passing `name` will fix all cases that are currently broken. this will just provide an error message ... however it wont let you know what the possible config names are (the user will still have to know which name they want)

PR #459, opened 2022-04-14
Optional Pubmed abstract downloader

HallerPatrick:
A utility function for conveniently downloading PubMed abstract for given pmids. This was discussed with @jason-fries and might come in handy.

This PR can be merged, when needed. Otherwise delete. :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ flake8>=3.8.3
 isort>=5.0.0
 aiohttp==3.8.1
 openpyxl>=3.0.9
+biopython==1.79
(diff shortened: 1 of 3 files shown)


jason-fries:
Hi @HallerPatrick 
Thanks for putting this together! We'll discuss during the QA/QC phase of the hackathon what we want to do with datasets that are fundamentally incomplete without some sort of join on another dataset like the PMC Open Subset. We'll need something like this tool, but it's not clear yet exactly what the right abstraction will look like.  I'll go ahead and close this PR for now.

PR #466, opened 2022-04-15
WIP: A dataloader to automatically check for data types, loads sample data and checks for different splits

debajyotidatta:
Hi, I wrote a simple app that you can run for QA to check for data types, some sample data and for different data splits. 

We can initially use it for the QA pass to check schema formatting and stuff, but eventually can remove some of these for making a datasets viewer like tool for our biomedical work, with some specific things for source schema and bigbio schema.

Some screenshots:

Screen Shot 2022-04-15 at 7 40 00 AM

From the left you can select the dataset and schema type and it will fill the rest.

Screen Shot 2022-04-15 at 7 40 16 AM

Displays explicit datatypes that we want to be consistent about.

Screen Shot 2022-04-15 at 7 40 27 AM

Finally loads data. (Only works for downloadable data.) I will make some enhancements to this, like giving the path to the local directory.

This was meant to be something quick so we can start the QA but feel free to suggest any features that might be helpful and I can add them.


app.py
```diff
@@ -0,0 +1,89 @@
...
+
+    if isinstance(features, datasets.features.Sequence):
+        return {"[]": render_features(features.feature)}
+    return features
+
+
+def load_ds(dataset_name):
+    path = f"biodatasets/{dataset_name}/{dataset_name}.py"
+    module_path = datasets.load.prepare_module(path, dataset=True)
+    builder_cls = datasets.load.import_main_class(module_path[0], dataset=True)
```
galtay, line 48: streamlit is so cool ... was trying this out and got an error b/c I have `datasets >= 2.0` ... seems they deprecated `prepare_module` ... this is the new way, 
```
path = f"biodatasets/{dataset_name}/{dataset_name}.py"
module = datasets.load.dataset_module_factory(path, dataset=True)
builder_cls = datasets.load.import_main_class(module.module_path, dataset=True)
```

galtay, line 48: streamlit is so cool ... was trying this out and got an error b/c I have `datasets >= 2.0` ... seems they deprecated `prepare_module` ... this is the new way, 
```
path = f"biodatasets/{dataset_name}/{dataset_name}.py"
module = datasets.load.dataset_module_factory(path, dataset=True)
builder_cls = datasets.load.import_main_class(module.module_path, dataset=True)
```

hakunanatasha:
Might be mistaken but deprecated from changes now moving to hub; can reopen if this is of interest.

PR #470, opened 2022-04-15
hacky hacky to test one config at a time

galtay:
This is a total quick hack to be able to test one config by name at a time.


galtay:
going to merge this, but we can remove later.

hakunanatasha:
@galtay in cases like this, it may be better to push the branch and allow people to just call the branch to run tests.

PR #476, opened 2022-04-16
Fix referenced ids check

leonweber:
Fixes a bug that made test for existence of referenced ids auto pass.


sg-wbi (approved):
Good catch @leonweber !

PR #489, opened 2022-04-19
Changed BioASQ Task B dataset name to match that of other BioASQ tasks.

jason-fries:
Some simple changes to naming conventions for the BioASQ Task B dataset.


galtay (approved):
looks like a sensible name change. LGTM

PR #492, opened 2022-04-20
Fix dataloader for Chia dataset

mariosaenger:
Hi,

this PR fixes the description and the entity references for the Chia dataset. 

Best,
Mario


leonweber (approved):
Thanks for the fixes @mariosaenger 🎉 . LGTM

PR #514, opened 2022-04-24
Minimal make package

galtay:
create a pip installable package called `biomed_datasets` by adding, 
* pyproject.toml
* setup.cfg
* setup.py

this allows anyone to `pip install -e .` and get the `biomed_datasets` package into their python environment
package requirements are split between install, dev, and build

* install: allows use of the package for datasets loading
* dev: adds linting/formatting packages
* build: adds packages for building and deploying to pypi

to try out the branch, create a new python environment and  
```
git checkout upstream minimal_make_package
pip install -e ".[dev,build]"
```

note that the above command should be given in the same directory as setup.cfg
to test it out try, 

```python
import biomed_datasets
print(biomed_datasets.__version__)
```

to build the package (in preparation for possible deployment to PyPI with twine)
```
python -m build
```

moves the `utils` directory inside the package
replace all imports of utils with imports of biomed_datasets


hakunanatasha (approved):
@galtay wonderful work.
I checked each file, and also made a testing script that loads every single dataset (including local, keeping track of which fails).

Amusingly, this approach seems to load more datasets than the current one. I am not sure why; the old datasets seems to run into issues with:
`
'ebm_pico',
'gnormplus',
'mantra_gsc',
'meddocan',
'ncbi_disease',
'pubtator_central',
'thomas2011'
`

Before merging, let's please finish the last few PRs; the older PRs will inherit the structure from before which may require some minor tweaks. I suspect it may be easy to merge new datasets into this PR.

Lastly, while the package setup is useful, we should think about what core classes + functions we want to expose as accessible (i.e. a form of load_datasets with a task-type input).

hakunanatasha:
@galtay  last note -prior to merging, let's make sure no one is still working on submitting a PR; the change to main will result in a pretty sizable conflict for a participant.

galtay:
thanks for reviewing @hakunanatasha ! 
you are correct, the vast majority of changes had to do with changing imports 
I don't think we have any open developer PRs that will be hugely hurt by this (or at least they will also just need an update to their imports)
Totally agree we could expose some extra functions like customized load_dataset functionality. maybe in another PR though? 
I was also thinking about the existing PRs but they keep rolling in and I think there will be some open no matter when we merge this. We could make an announcement in discord asking people who are working on open PRs  to be careful pulling from master while their PRs are open. 
I'm happy to fixup the imports in open PRs after they are merged so that they can use the current imports and tests (if that makes sense)

galtay:
closing in favor of #520

PR #526, opened 2022-04-28
BigBio Meta-dataloader

sg-wbi:
This is a just a tow example to propose the idea of having a single access point for all datasets. 

Like the `transformers` library has a `Trainer` class we should have something similar that offers a customizable access to a subset (or all) of the datasets, e.g. by task, by benchmark (BLEURB)...


sg-wbi:
I'll open a new one after the package PR is merged

PR #529, opened 2022-04-30
unify tests after bigbio package

galtay:
first attempt at unifying and cleaning up test script. 

there are only 3 args to pass in now, 
* path = the path to the dataloader script
* data_dir (only needed for local datasets) 
* config_name (only needed to test a single config by name)

by default all configs defined are run thru the unit tests. 
the diff is a mess to look at so it might be easier to view the full file or check out the branch and run the tests. 

e.g. 
`python -m tests.test_bigbio biodatasets/an_em/an_em.py`
`python -m tests.test_bigbio biodatasets/n2c2_2011/n2c2_2011.py --data_dir ~/data/big_science_biomedical/n2c2_2011`


galtay:
replacement for #527

PR #524, opened 2022-04-28
[WIP] Adversarial 'meta' unit test 

sg-wbi:
This includes a `dummy_features` file containing (hopefully) easy to tweak features to verify each "subtest" of the main test file.

An example is provided with `dummy_kb_dataloader.py` which implements an example which fails the "global_unque_ids" test.

Feedback is greatly appreciated.

If this is this format is suitable I;ll add more "tests" for tests

PR #534, opened 2022-05-02
Dummy dataloader v2

sg-wbi:
Supersedes #524 

It adds a `DummyKBExample` to reduce boilerplate and sets examples per split, this way we can test the bypass args here #533.

An example is provided with dummy_kb_dataloader.py which implements an example which fails the "global_unque_ids" test.


hakunanatasha:
@sg-wbi looks like this is fine; to run, I did the following:
`python -m tests.test_bigbio tests/dummy/dummy_kb_dataloader.py` and saw a failure.

The generator is very clever. Can we add in the `tests` case one for every sub function? This should be

- passage offsets
- entity offsets
- event offsets
- coreference
- q+a
- text2text
- text labels

this is a lot of work so i dont mind if its a TODO but this is good to merge as is imo

hakunanatasha (approved):

hakunanatasha:
@sg-wbi are we merging this?

sg-wbi:
I am merging this as a first building block. I created this #665 so we do not forget about this.

PR #535, opened 2022-05-02
BigBioDataloader (meta-dataloader)

sg-wbi:
This is to re-open (#526 ) the discussion on having a single class as access point to all dataloaders with customizable configurations (e.g. load all BLEURB datasets).

### DIFF
--- a/bigbio/dataloader.py
+++ b/bigbio/dataloader.py
@@ -1,0 +1,120 @@
+##!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Meta dataloader to load "load configurations" for all available datasets according to `_SUPPORTED_TASKS`
+"""
+
+import importlib
+import os
+from collections import defaultdict
+from glob import glob
+from types import ModuleType
+from typing import Optional, Union
+
+import datasets
+from bibgio.utils.constants import Tasks
+
+_SCHEMA_TO_TASKS = {
+    "bigbio_kb": [
+        Tasks.NAMED_ENTITY_RECOGNITION,
+        Tasks.NAMED_ENTITY_DISAMBIGUATION,
+        Tasks.EVENT_EXTRACTION,
+        Tasks.RELATION_EXTRACTION,
+        Tasks.COREFERENCE_RESOLUTION,
+    ],
+    "bigbio_qa": [Tasks.QUESTION_ANSWERING],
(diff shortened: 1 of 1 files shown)


bigbio/dataloader.py
```diff
@@ -0,0 +1,323 @@
...
+    def __init__(
+        self,
+        tasks: Optional[list[Tasks]] = None,
+        data_dir: Optional[str] = None,
+        skip_huge: bool = True,
+        skip_weak: bool = True,
+        skip_local: bool = True,
+        exclude: Optional[list] = None,
+    ):
+        self.scripts = sorted(glob(os.path.join("biodatasets", "*", "*.py")))
```
galtay, line 77: this will only work from within the package but we'd like it to work anywhere someone is using the package. maybe try something like, 
```
path_to_here = pathlib.Path(__file__).parent.absolute()                                                                                                                                                                           
self.path_to_biodatasets = path_to_here / "../" / "biodatasets"                                                                        
self.scripts = sorted(glob(os.path.join(self.path_to_biodatasets, "*", "*.py"))) 
```

bigbio/dataloader.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+
+        bigbio_datasets = defaultdict(list)
+
+        for script in self.scripts:
+
+            if Reference(script).to_name() in self.exclude:
+                continue
+
+            module = self.load_module(script)
```
galtay, line 170: the logic in `load_module` will break if this is run outside the package (e.g. try running from your home directory)

galtay (changes requested):
some initial comments for review. will come back to this ... I have a sense that we are over engineering a bit here, but will come back with some more productive comments :)

bigbio/dataloader.py
```diff
@@ -0,0 +1,323 @@
...
+        "pubtator_central_bigbio_kb",
+    ],
+}
+
+
+# NOTE:
+# weak datasets are the unstructured ones,
+# i.e. not providing a clear task but being more of a resource,
+# e.g. PubTator and MIMICIII
+_WEAK: dict = {
```
galtay, line 55: maybe we should use a name like `_RESOURCE` here ?

galtay:
@sg-wbi you mind if I pick this work up from here?

sg-wbi:
Superseded by #561

PR #536, opened 2022-05-02
add _LOCAL variable

sg-wbi:
This adds a `_LOCAL` metadata variable to all `examples` (as a showcase). This should allow easier handling.

If this is fine, I;ll add it to all `biodatasets` and set it initially to `False`. Then (hopefully in a joint effort) we can set the attribute correctly


sg-wbi:
@hakunanatasha you were on board with this right?

examples/n2c2_2011.py
```diff
@@ -69,13 +69,15 @@
...
 from datasets import Features, Value
+
 from bigbio.utils import schemas
 from bigbio.utils.configs import BigBioConfig
 from bigbio.utils.constants import Tasks
 
 _DATASETNAME = "n2c2_2011"
 
 # https://academic.oup.com/jamia/article/19/5/786/716138
+_LOCAL = False
```
galtay, line 80: this is a local dataset

galtay:
I know n2c2_2011 is a local dataset and some of these others might be. Can you run the tests on each one and see what is local? You should get a readable error if you try and test a local dataset without using the data_dir and that should let you know which ones are local (or you can just grep for `raise ValueError` in the examples directory

galtay (changes requested):
check which are actually local

sg-wbi:
@galtay I have added the `_LOCAL` attribute with the correct value to all datasets in `examples` and `biodatasets`.
I used [this script](https://gist.github.com/sg-wbi/246aa8c7b542ef7d6f7318670c66ce19) to check if the datasets are local and this is the output:
[local.txt](https://github.com/bigscience-workshop/biomedical/files/8620167/local.txt) (the file is actually in JSON format but github does not allow the upload of .json)

hakunanatasha:
@galtay @sg-wbi surprised; all n2c2 should be local

hakunanatasha:
@sg-wbi @galtay I ran each script for failures (just a simple load_check); I labelled the ones I could see are explicitly local by monitoring both present URLs and strange situations where a zip file is passed with no real URL path to a downloadable website.

sg-wbi:
> surprised; all n2c2 should be local

And they are not?

```bash
$ rg "n2c2.*.py" local.txt
62: "biodatasets/n2c2_2006_deid/n2c2_2006_deid.py": [
68: "biodatasets/n2c2_2006_smokers/n2c2_2006_smokers.py": [
74: "biodatasets/n2c2_2008/n2c2_2008.py": [
80: "biodatasets/n2c2_2009/n2c2_2009.py": [
86: "biodatasets/n2c2_2010/n2c2_2010.py": [
92: "biodatasets/n2c2_2011/n2c2_2011.py": [
98: "biodatasets/n2c2_2018_track1/n2c2_2018_track1.py": [
104: "biodatasets/n2c2_2018_track2/n2c2_2018_track2.py": [
```

``` bash
ls . | rg "n2c2"
 n2c2_2006_deid
 n2c2_2006_smokers
 n2c2_2008
 n2c2_2009
 n2c2_2010
 n2c2_2011
 n2c2_2018_track1
 n2c2_2018_track2
```

galtay (approved):
looks good to me.we can fix things that end up being wrong if any. the ones I know are local look good and the methodology for determining what is local seems good too. I'd like to merge before any get too out of date.

Issue #155, opened 2022-02-26
Create dataset loader for UMNSRS Reference Standard


galtay:
- **Name:** UMNSRS Reference Standard
- **Description:** UMNSRS, developed by Pakhomov, et al., consists of 725 clinical term pairs whose semantic similarity and relatedness. 
- **Task:** Semantic Similarity
- **Paper:** https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3041430/
- **Data:** https://nlp.cs.vcu.edu/data.html#umnsrs
- **License:** ?
- **Motivation:** Evaluation

trishalaneeraj:
#self-assign

trishalaneeraj:
Still working on this!

jason-fries:
Thanks!

PR #538, opened 2022-05-03
Fix dataloader for umnsrs

nomisto:
Rework of dataloader #155:

  Why: Unit tests failed - The original source could not be programatically downloaded with `download_and_extract` since the it the url automatically redirected to a website for accepting cookies
- Added subset "relatedness"
  Original dataloader implemented the similarity dataset (e.g., whale-dolphin) but not the relatedness dataset (e.g. needle-thread)
- Added subsets "mod"
  There is a modified set for "similarity" and "relatedness" which 'exclude control samples and those pairs that did not match text in clinical, biomedical and general English corpora.', see ref paper in https://conservancy.umn.edu/handle/11299/196265
- Added license

### DIFF
--- a/biodatasets/umnsrs/umnsrs.py
+++ b/biodatasets/umnsrs/umnsrs.py
@@ -86,7 +86,7 @@ _SOURCE_VERSION = "1.0.0"
 _BIGBIO_VERSION = "1.0.0"
 
 
-class MayosrsDataset(datasets.GeneratorBasedBuilder):
+class UmnsrsDataset(datasets.GeneratorBasedBuilder):
     """UMNSRS, developed by Pakhomov, et al., contains clinical term pairs whose semantic similarity and
     relatedness were scored by experts."""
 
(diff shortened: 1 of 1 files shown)


biodatasets/umnsrs/umnsrs.py
```diff
@@ -48,55 +49,74 @@
...
-    Tasks.SEMANTIC_SIMILARITY
-]  # example: [Tasks.TRANSLATION, Tasks.NAMED_ENTITY_RECOGNITION, Tasks.RELATION_EXTRACTION]
+_SUPPORTED_TASKS = [Tasks.SEMANTIC_SIMILARITY]
 
 _SOURCE_VERSION = "1.0.0"
 
 _BIGBIO_VERSION = "1.0.0"
 
 
 class MayosrsDataset(datasets.GeneratorBasedBuilder):
```
leonweber, line 89: The class is called MayosrsDataset is that correct?

leonweber (changes requested):

biodatasets/umnsrs/umnsrs.py
nomisto: Ah no, good catch!

galtay (approved):

galtay (approved):

Issue #161, opened 2022-02-27
Create dataset loader for MayoSRS Reference Standard


galtay:
- **Name:** MayoSRS Reference Standard
- **Description:** MayoSRS consists of 101 clinical term pairs whose relatedness was determined by nine medical coders and three physicians from the Mayo Clinic. 
- **Task:** Semantic Similarity
- **Paper:** https://pubmed.ncbi.nlm.nih.gov/16875881/
- **Data:** https://nlp.cs.vcu.edu/data.html#mayosrs
- **License:** ?
- **Motivation:**

trishalaneeraj:
#self-assign

jason-fries:
Hi @trishalaneeraj  can you let us know if you are still working on this so we can update our project board? Please just notify us the status **by Friday April 8**.  You can response to this comment or ping us on Slack or Discord. 

No worries if you are not finished but still intend to work on this!

trishalaneeraj:
Hi @jason-fries, I'm still working on this

PR #539, opened 2022-05-03
Fix dataloader for mayosrs

nomisto:
Rework of dataloader #161:

  Why: Unit tests failed - The original source could not be programatically downloaded with `download_and_extract` since the it the url automatically redirected to a website for accepting cookies
- Added license

### DIFF
--- a/biodatasets/mayosrs/mayosrs.py
+++ b/biodatasets/mayosrs/mayosrs.py
@@ -141,7 +141,7 @@ class MayosrsDataset(datasets.GeneratorBasedBuilder):
                 for id_, row in data.iterrows():
                     yield id_, {
                         "id": id_,  # uid is an unique identifier for every record that starts from 1
-                        "document_id": "NULL",
+                        "document_id": id_,
                         "text_1": row["text_1"],
                         "text_2": row["text_2"],
                         "label": str(row["label"]),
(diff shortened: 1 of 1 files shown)


sg-wbi:
Hey @nomisto great work catching this! When I try to load the dataset I get this error:
```python
File ~/.cache/huggingface/modules/datasets_modules/datasets/mayosrs/7e10c2a67d74a2869f076925fb3525f08252a281ef3c692445129d26a87dc457/mayosrs.py:152, in MayosrsDataset._generate_examples(self, filepath, split)
    148     csv_reader_code = csv.reader(
    149         csv_file, quotechar='"', delimiter=",", quoting=csv.QUOTE_ALL, skipinitialspace=True
    150     )
    151     for id_, row in enumerate(csv_reader_code):
--> 152         label, code1, code2 = row[0].split("<>")
    153         data_code.append([label, code1, code2])
    155 data_texts = []

ValueError: not enough values to unpack (expected 3, got 1)
```
Is it the same one you are fixing with this PR?

nomisto:
Yes! So basically the reason for this is that it does not download the correct resource but a HTML file (the host automatically redirects to a "accept cookies" page if a specific cookie is not set).

The same problem (same host) is with datasets Minimayosrs and umnsrs (See other PRs)

sg-wbi (approved):
I confirm that this fixes the loading issue and all tests pass. Thanks again @nomisto, good catch!

Issue #162, opened 2022-02-27
Create dataset loader for MiniMayoSRS Semantic Similarity Reference Standard


galtay:
- **Name:** MiniMayoSRS Semantic Similarity Reference Standard
- **Description:** MiniMayoSRS is a subset of the MayoSRS and consists of 30 term pairs on which a higher inter-annotator agreement was achieved. The average correlation between physicians is 0.68. The average correlation between medical coders is 0.78.
- **Task:** Semantic Similarity
- **Paper:** https://pubmed.ncbi.nlm.nih.gov/16875881/
- **Data:** https://nlp.cs.vcu.edu/data.html#mayosrs
- **License:** ?
- **Motivation:**

trishalaneeraj:
#self-assign

jason-fries:
Hi @trishalaneeraj  can you let us know if you are still working on this so we can update our project board? Please just notify us the status **by Friday April 8**.  You can response to this comment or ping us on Slack or Discord. 

No worries if you are not finished but still intend to work on this!

trishalaneeraj:
Hi @jason-fries, I'm still working on this

debajyotidatta:
Hi @trishalaneeraj : This looks great! But in the source schema do you mind changing the values that are string to feature type float? It would be nice to have the float as the original source schema since that was intended.

nomisto:
I am currently doing QA for this dataset and I ran into several issues:

1. Unit test fail because the data could not be downloaded (Would be fixed with #532)
2. The source data seems to have several flaws which result in wrong examples
    1. [Term Pairs](https://nlp.cs.vcu.edu/data/similarity-data/MiniMayoSRS.terms) and [physician judgments](https://nlp.cs.vcu.edu/data/similarity-data/MiniMayoSRS.physicians) have different sizes (30 vs 29)
    2. The indices of "Term Pairs" and "physician judgments" are not the same. f.e. the sample of `a["train"][28]` is `{'text_1': 'Varicose vein', 'text_2': 'Entire knee meniscus', 'label_physicians': 1.0, 'code_1_physicians': 'C0020473', 'code_2_physicians': 'C0027627', 'label_coders': 1.0, 'code_1_coders': 'C0020473', 'code_2_coders': 'C0027627'}`, however CUI C0020473 is Hyperlipidemia and C0027627 Tumor metastasis
3. There is no need to have  code_1_physicians and code_1_coders; they HAVE TO be the same (same goes for code_2)

[...]

PR #540, opened 2022-05-03
Fix dataloader for minimayosrs

nomisto:
Rework of dataloader #162:

  - Why: Unit tests failed - The original source could not be programatically downloaded with `download_and_extract` since the it the url automatically redirected to a website for accepting cookies
  - Furthermore, the source data seems to have several flaws which result in wrong examples
    - Term Pairs and physician judgments have different sizes (30 vs 29)
    - The indices of "Term Pairs" and "physician judgments" are not the same. f.e. the sample of a["train"][28] is {'text_1': 'Varicose vein', 'text_2': 'Entire knee meniscus', 'label_physicians': 1.0, 'code_1_physicians': 'C0020473', 'code_2_physicians': 'C0027627', 'label_coders': 1.0, 'code_1_coders': 'C0020473', 'code_2_coders': 'C0027627'}, however CUI C0020473 is Hyperlipidemia and C0027627 Tumor metastasis
- There is no need to have code_1_physicians and code_1_coders; they HAVE TO be the same (same goes for code_2), there is now only `code_1` and `code_2`
- Label in bigbio config is now avg of the two `str((row["label_physicians"] + row["label_coders"]) / 2)` (similar to biosses) rather than just physician
- Added license

### DIFF
--- a/biodatasets/minimayosrs/minimayosrs.py
+++ b/biodatasets/minimayosrs/minimayosrs.py
@@ -18,10 +18,11 @@ MayoSRS consists of 101 clinical term pairs whose relatedness was determined by
 nine medical coders and three physicians from the Mayo Clinic.
 """
 
-import csv
 from typing import Dict, List, Tuple
 
 import datasets
+import pandas as pd
+
 from bigbio.utils import schemas
 from bigbio.utils.configs import BigBioConfig
 from bigbio.utils.constants import Tasks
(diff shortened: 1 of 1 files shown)


sg-wbi (approved):
Thank you for fixing this too @nomisto! I confirm that now the dataset loads properly and passes the tests.

PR #546, opened 2022-05-05
Fix LLL: correct entity definition, use BigBio.NULL instead of None

leonweber:
Two fixes for LLL:
* before all words were entities in BigBio.KB, changed to actual biological entities
* use BigBio.NULL instead of None to make tests happy


galtay (approved):

galtay:
fyi i ran the tests like so, 
```
python -m tests.test_bigbio biodatasets/lll/lll.py --bypass_split_key_pairs test,relations test,entities
```

PR #550, opened 2022-05-06
Fix bionlp_st_2013_ge

sg-wbi:
It seems that the pattern for the failed check on referenced ids is because of missing types in `_ENTITY_TYPE` for the brat parser.
However, this time if I try to add all the entity type associated to a `T` line I get this error:
```python
Traceback (most recent call last):
  File "/home/ele/Desktop/projects/tm/bigbio/tests/test_bigbio.py", line 133, in runTest
    self.dataset = datasets.load_dataset(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/load.py", line 1691, in load_dataset
    builder_instance.download_and_prepare(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 605, in download_and_prepare
    self._download_and_prepare(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 1104, in _download_and_prepare
    super()._download_and_prepare(dl_manager, verify_infos, check_duplicate_keys=verify_infos)
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 694, in _download_and_prepare
    self._prepare_split(split_generator, **prepare_split_kwargs)
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/datasets/builder.py", line 1087, in _prepare_split
    for key, record in logging.tqdm(
  File "/home/ele/.venv/bigbio/lib/python3.9/site-packages/tqdm/std.py", line 1180, in __iter__
    for obj in iterable:
  File "/home/ele/.cache/huggingface/modules/datasets_modules/datasets/bionlp_st_2013_ge/49256844b3138a792d795859ff96d8ff12db17f1ca68c748dcfa8bccdda7cdb9/bionlp_st_2013_ge.py", line 250, in _generate_examples
    example = parsing.brat_parse_to_bigbio_kb(
  File "/home/ele/Desktop/projects/tm/bigbio/bigbio/utils/parsing.py", line 347, in brat_parse_to_bigbio_kb
    trigger = id_to_event_trigger[event["trigger"]]
KeyError: 'T5'

```

### DIFF
--- a/biodatasets/bionlp_st_2013_ge/bionlp_st_2013_ge.py
+++ b/biodatasets/bionlp_st_2013_ge/bionlp_st_2013_ge.py
@@ -94,7 +94,28 @@ class bionlp_st_2013_ge(datasets.GeneratorBasedBuilder):
 
     DEFAULT_CONFIG_NAME = "bionlp_st_2013_ge_source"
 
-    _ENTITY_TYPES = {"Protein", "Entity"}
+    # _ENTITY_TYPES = {
+    #     "Protein",
+    #     "Entity",
+    # }
+    _ENTITY_TYPES = [
+        "Phosphorylation",
+        "Acetylation",
+        "Localization",
+        "Ubiquitination",
+        "Anaphora",
+        "Transcription",
+        "Protein_catabolism",
+        "Gene_expression",
+        "Negative_regulation",
+        "Regulation",
+        "Protein_modification",
+        "Protein",
+        "Entity",
+        "Positive_regulation",
+        "Binding",
+        "Deacetylation",
+    ]
 
     def _info(self):
         """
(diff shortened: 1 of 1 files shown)


sg-wbi:
@leonweber @barthfab any idea?

sg-wbi:
By looking at `parse_brat` function, it seems that there is kind of a conflict, i.e. a `T` line cannot be both a text bound annotation and a event trigger. However, if we do not include all types associatged to a `T` line we loose annotations and hence get the failure in the tests. 

Or I am completely misunderstanding this... xD

leonweber:
The general idea of BRAT is that there is no distinction between entities and event triggers, both are simply annotated as text bound annotations (T lines). For the BigBio schema we need to figure out which Ts correspond to entities and which to events. We do this by manually providing the list of entity types to the parser which then assumes that every T with a type not in the entity list is an event trigger. 

I think that your diagnosis is correct and the reference errors come from missing entity types in this list. However, we cannot fix it by using all possible types, because then we would again lose the distinction between entities and event triggers. We have to check this on a case by case basis.

For this specific dataset, the issue seems to be that _Anaphora_ is not in the entity list. This makes sense, because it is not a biological entity. This is a special case, because all relations seem to be coreference annotations. I think we should include anaphoras in the entity list and transform the relations to coreference annotations. We could do this by using networkx do build a relation graph (entities/anaphoras are nodes and  relations are edges) and then using all disconnected subgraphs as coreference sets. What do you think?

sg-wbi:
Fixed with #553

PR #553, opened 2022-05-09
change brat_parse_to_bigbio_kb

barthfab:
Changing the brat_parse_to_bigbio_kb to fix issue #550. 
Test results of all data parser that use the brat parser is attached (except of pcr): [output.txt](https://github.com/bigscience-workshop/biomedical/files/8662459/output.txt)


bigbio/utils/parsing.py
```diff
@@ -353,10 +345,18 @@ def brat_parse_to_bigbio_kb(brat_parse: Dict, entity_types: Iterable[str]) -> Di
...
+    anno_ids = [ref_id['id'] for ref_id in non_event_ann]
+    for ann in non_event_ann:
+        entity_ann = ann.copy()
+        entity_ann["id"] = id_prefix + entity_ann["id"]
+        entity_ann["normalized"] = ref_id_to_normalizations[ann["id"]]
+        unified_example["entities"].append(entity_ann)
     # massage relations
     unified_example["relations"] = []
     for ann in brat_parse["relations"]:
+        if ann["head"]["ref_id"] not in anno_ids or ann["tail"]["ref_id"] not in anno_ids:
```
sg-wbi, line 358: Can we have a `logger.warning` here? Saying:
```python
"`bigbio_kb` schema allows for coreference relations only between entities"
```

sg-wbi:
There was a lot of background discussion on this. In summary this changes how brat files are parsed: 
1. instead of having to provided a list of entity (`_ENTITY_TYPE`) to discriminate between `entities` and event `trigger`s, the former are defined by not being `trigger`s, i.e. first parse all events to collect `trigger`s then define `text_bound_annotations`.
2. As of now the `bigbio_kb` schema allows for coreference relations only between `entities`, while in some datasets (e.g. `bionlp_st_2013_gro`) there can be coreference relations between `entities` and `events`. For now these will be skipped (i.e. not included in the dataset). (This will change in a v2 schema?)

sg-wbi:
None of the datasets which uses this function break with these changes so to me it's ready to be merged, once we add that little warning just so we do not forget we are doing (2).

@leonweber if you are fine with this I'll merge, this way we have all the event extraction datasets working.

sg-wbi:
I'll open a separate PR to do a clean-up on all brat datasets since `_ENTITY_TYPE` is no longer required (for good IMO).

leonweber (approved):
LGTM. Thanks for your efforts here @barthfab and @sg-wbi !

sg-wbi (approved):
All the credit to @barthfab: he spotted the true source of error! Thanks!

barthfab:
The results of the latest test run using https://github.com/bigscience-workshop/biomedical/pull/564 

[output.txt](https://github.com/bigscience-workshop/biomedical/files/8670454/output.txt)

PR #559, opened 2022-05-10
Fix source config of gnormplus

nomisto:
Could not load source config of gnormplus due to the attribute `passage.total_span.length` not found. Maybe this is due to an bioc version upgrade? Fixed it by calculating the length of the span "manually".


hakunanatasha:
@nomisto are you using the bio-c version that corresponds to the req. file? Just let us know and if so i can merge.

nomisto:
Oh sorry, yes i've used an old version of bioc. Closing this now

PR #563, opened 2022-05-11
remove `_` prefix from constants

sg-wbi:
A couple of constants were not renamed after #560


sg-wbi:
changed `_VALID_TASKS` and `_SCHEMA_TO_FEATURES` to values `constants`

PR #581, opened 2022-05-13
format,recurse,rename

galtay:
* add a list of broken configs so they wont get loaded 
* rename so we just have two  classes `BigBioConfigHelper` and `BigBioConfigHelpers`
* ^ is the acknowledgement that all we are doing is manipulating and filtering these configs 
* make some of the more opaque attributes underscore attributes and pull out the important pieces in understandable names (e.g. `py_mod` -> `_py_mod` and `py_mod._LICENSE` -> `license`  
* make `BigBioConfigHelpers` iterable instead of always referring to the attached list
* make most methods return another instance of `BigBioConfigHelpers` (hello from your friendly recursion person) 
* ^ allows things like chaining filtered calls (e.g. `filtered_helpers = helpers.filtered(lambda x: filter_1).filtered(lambda x: filter 2)...`

### DIFF
--- a/bigbio/dataloader.py
+++ b/bigbio/dataloader.py
@@ -7,9 +7,10 @@ import logging
 import os
 import pathlib
 from types import ModuleType
-from typing import Callable, List, Optional
+from typing import Callable, Iterable, List, Optional
 
 from dataclasses import dataclass
+from dataclasses import field
 import datasets
 from datasets import load_dataset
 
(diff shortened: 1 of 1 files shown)


galtay:
https://colab.research.google.com/drive/1YwJbpfh4mzGpSAntgBOEsDiNds99O3GV?usp=sharing

galtay:
```python
from bigbio.dataloader import BigBioConfigHelpers
conhelps = BigBioConfigHelpers()
tmvar_datasets = [                                                                                     
    load_dataset(**helper.get_load_dataset_kwargs())                                                   
    for helper in (
        conhelps
        .filtered(lambda x: "tmvar" in x.dataset_name)
        .filtered(lambda x: x.is_bigbio_schema)                                   
    )                                                        
] 
```

PR #585, opened 2022-05-17
Fix test logging multilabel check

sg-wbi:
Logging of tests for multilabel `db_id`/`feature_type` were raising an error: message was not parsed correctly.


hakunanatasha (commented):
Comments via slack; some major questions:

(1) Why is 2 errors acceptable? The logic breaks at 2
(2) Might be useful to have an error at the break statements on where the code exists + why it reached 2 issues.
(3) May be simpler to just do `while len(warnings.values()) < 2`

sg-wbi:
> (1) Why is 2 errors acceptable? The logic breaks at 2
(2) Might be useful to have an error at the break statements on where the code exists + why it reached 2 issues.
(3) May be simpler to just do while len(warnings.values()) < 2

The idea was that two connectors should warrant a cleaning pass for the dataset. But it makes no sense, since the user does not know what connectors she should look for. I removed this. 

Now you get one warning per connector type

galtay (approved):

PR #600, opened 2022-05-20
Multilabel db check

sg-wbi:
Include `db_name` field in the multilabel test.


sg-wbi:
@galtay as per your suggestion. At least I think this is what you meant!

galtay:
@sg-wbi yes, thank you!

galtay (approved):

PR #603, opened 2022-05-20
Normalize licenses

sg-wbi:
The idea is to have classes for different type of licenses whit each license having a subset of available parameters: name, type, version, link, description.

This entails replacing:
```python
_LICENSE = "CC BY-NC-SA"
_LICENSE= "CC BY-NC-SA 3.0"
```
With

```python
from bigbio.utils import license
_LICENSE = license.CreativeCommons(type="BY-NC-SA")
_LICENSE = license.CreativeCommons(type="BY-NC-SA", version=3.0)
``` 
and
```
return datasets.DatasetInfo(
            description=_DESCRIPTION,
            features=features,
            homepage=_HOMEPAGE,
            license= str(_LICENSE),
            citation=_CITATION,
        )
```
Special ones are `Custom` for dataset-specific license and `PubliclyAvailable` for those datasets which can be freely downloaded but do not provide license information.


sg-wbi:
@jason-fries @hakunanatasha @galtay what do you think?

jason-fries:
Hey @sg-wbi 
This is great! What are your thoughts on making this a generate attribute mixin so we can define structured/enum type values for other information (languages, subtasks, etc)

galtay:
In general, I like the idea of formalizing the license fields ... I need to think a bit more about the structures in this PR (e.g. not sure we need a distinct class for each license type) but I'll post more on that after I have some ideas :) 
Eventually it would be nice if this field in the `BigBioConfigHelper` was an instance of whatever we come up with here 
https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/dataloader.py#L101

bigbio/utils/license.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __str__(self):
+        """
+        Override to get a formatted full description of the license
+        """
+        return self.to_string()
+
+
+@dataclass
+class CreativeCommons(LicenseMixin):
```
galtay, line 53: do you imagine "mixing in" more than one class here? just an idea, but could we get away with a single `LICENSE` dataclass here and create instances of it that we can use elsewhere? 
e.g. 
```
CREATIVE_COMMONS_BY_NC_SA_4p0 = LICENSE(
  name = "CC BY-NC-SA" # also maybe chosen from an enum
  family = "CC" # maybe chosen from an enum
  version = "4.0"
  )
```

jason-fries:
@sg-wbi @galtay 
At a basic level, I think we need some sort of `DatasetAttribute` class that wrangles categories (`License`, `Language`, etc) of enumerated types. This should be super simple for the first version and primarily handle the error mode highlighted by the proliferation of license strings used by hackathon participants. This will also let us generate attribute counts for the paper. 

That said, I can definitely imagine more sophisticated functionality being useful here beyond enums. For example with licenses, I could imagine wanting more information available during querying (commercial use permitted, ShareAlike permitted, etc).

sg-wbi:
@jason-fries 
> At a basic level, I think we need some sort of DatasetAttribute class that wrangles categories (License, Language, etc) of enumerated types. This should be super simple for the first version and primarily handle the error mode highlighted by the proliferation of license strings used by hackathon participants.

I get your point however I think that for metadata s.a. `Language` a simple `Enum` should be sufficient since I do not see what other functionalities might be shared w/ `License` for instance. 

>  I could imagine wanting more information available during querying (commercial use permitted, ShareAlike permitted, etc).

I really like this idea :) This is one more reason to have a `License` class. 

>  type values for other information (languages, subtasks, etc

What do you have in mind w/ `subtasks`?

sg-wbi:
@galtay 
> do you imagine "mixing in" more than one class here? just an idea, but could we get away with a single LICENSE dataclass here and create instances of it that we can use elsewhere?
e.g.

You are right here, we can get away w/ a single class. I'll make a new PR for this.

sg-wbi:
@galtay @jason-fries @hakunanatasha 

This made me also think we are starting to have multiple datasets attributes/metadata (e.g. `_LOCAL`,  `_PUBMED`, `_LANGS`) and we catch that these are missing only downstream in `BigBioConfigHelper` (which complains if it does not find them).

How about having a 

```python
@dataclass
DataloaderCard:
   license : License
   langs : constants.Lang
   local : bool
   is_pubmed: bool
```
in the dataloaders script which requires all metadata at instantiation? This way someone implementing a dataset is forced to fill in all metadata.

galtay:
> @galtay @jason-fries @hakunanatasha
> 
> This made me also think we are starting to have multiple datasets attributes/metadata (e.g. `_LOCAL`, `_PUBMED`, `_LANGS`) and we catch that these are missing only downstream in `BigBioConfigHelper` (which complains if it does not find them).
> 
> How about having a
> 
> ```python
> @dataclass
> DataloaderCard:
>    license : License
>    langs : constants.Lang
>    local : bool
>    is_pubmed: bool
> ```
> 
> in the dataloaders script which requires all metadata at instantiation? This way someone implementing a dataset is forced to fill in all metadata.

In theory I like that idea. In practice, I think we will keep on finding things we want to query and then having to add them into that `DataloaderCard` class and we will constantly have some datasets that have certain attributes filled in and others that don't ... Having said that, I think the case is pretty clear for `license`, `langs`, and `local` ... although I can think of at least one where `lang` applies at the config level and not the dataset level (e.g. muchmore). BUT I can also see this dataloader card langs meaning "all languages present in the dataset". And then there are always things like "is in BLURB" which definitely apply at the config level (e.g. bioasq task 7b bigbio kb schema) and probably? should remain as a collection of configs specified outside to the dataset loader scripts themselves.

galtay:
p.s. found this nice list in the `datasets` repo https://github.com/huggingface/datasets/blob/bb42d5c8813e12df44ddf527b08752e8047458f7/src/datasets/utils/resources/licenses.json

sg-wbi:
> In practice, I think we will keep on finding things we want to query and then having to add them into that DataloaderCard class and we will constantly have some datasets that have certain attributes filled in and others that don't

Yeah you are right, it come to me the second after I finished writing...

> although I can think of at least one where lang applies at the config level and not the dataset level (e.g. muchmore). BUT I can also see this dataloader card langs meaning "all languages present in the dataset".

Let's talk about this here #619

> And then there are always things like "is in BLURB" which definitely apply at the config level (e.g. bioasq task 7b bigbio kb schema) and probably? should remain as a collection of configs specified outside to the dataset loader scripts themselves.

Yeah for this use case I also think it should be outside the dataloader, i.e. having a constant saying "if you want BLURB go pick this and that dataloader with this and that config".

> I like the dataclass as a "TODO" but I think we can accomplish all of its functionality with _ variables and checking for their existence in the tests right? or were you thinking something like loading the dataset would be prevented if they were not filled out?

The second one :) But I also agree that doing this in the test is fine.

> p.s. found this nice list in the datasets repo https://github.com/huggingface/datasets/blob/bb42d5c8813e12df44ddf527b08752e8047458f7/src/datasets/utils/resources/licenses.json

Very nice finding! I am using this in #619. Any comment appreciated.

sg-wbi:
Closing in favor of #620

PR #610, opened 2022-05-21
add set of blurb configs

galtay:
@sg-wbi @ruisi-su 

looks like we are only missing GAD. 
I wasn't positive that gnormplus was BC2GM but I think it is.


sg-wbi:
>  I wasn't positive that gnormplus was BC2GM but I think it is.

Yes gnormplus is (an updated) BC2GN.

Issue #596, opened 2022-05-20
bc5cdr normalizations have -1s 


galtay:
https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/biodatasets/bc5cdr/bc5cdr.py#L222

I think we want something else here.

PR #611, opened 2022-05-21
use empty lists instead of -1 for missing normalizations

galtay:
@jason-fries @leonweber 
This is my quick fix ... let me know if you see any big problems.


galtay:
fixes #596

PR #619, opened 2022-05-23
Add _LANGUAGES metadata

sg-wbi:
- create a `Lang` enum with languages codes, loaded from json file `bigbio/utils/resources/languages.json`
- add `_LANGUAGES` module constant for each dataset 

@galtay w.r.t the point you mentioned in #603:

> although I can think of at least one where lang applies at the config level and not the dataset level (e.g. muchmore). BUT I can also see this dataloader card langs meaning "all languages present in the dataset".

I think that for now we want case (2), i.e. this dataset offers the following languages, as we do with `_SUPPORTED_TASK`, but I agree that there are nuances to this and we should be able to pull out the exact configs by language.


galtay (approved):
wow, lots of edits, nice work!

PR #620, opened 2022-05-23
Normalize license v2

sg-wbi:
This creates a `Licenses` enum with licenses loaded from json file `bigbio/utils/resources/licenses.json` (thanks @galtay for finding this!!!)

This allows to have all licenses available in one class. 

The additional nifty trick is that the values in the `Enum` are instances of a `License` class, which allows to define custom properties, e.g. we can do:

```python
[ins] In [4]: from bigbio.utils.license import Licenses
[ins] In [5]: Licenses.CC_BY_SA_1p0.is_share_alike
Out[5]: False
```
(yeah the result is wrong, it's just an example)

This includes only one `License` subclass, i.e. `CustomLicense` for edge-cases w/ custom licenses e.g. "Genia project".


bigbio/utils/license.py
```diff
@@ -0,0 +1,116 @@
...
+@dataclass
+class License:
+    """
+    Base class from which all licenses inherit
+    """
+
+    name: Optional[str] = None
+    text: Optional[str] = None
+    link: Optional[str] = None
+    version: Optional[float] = None
```
galtay, line 25: even if its not always filled out, I was thinking it would be nice to have a `provenance` field (or some name like that) that could be used to indicate the source of the license attribution (i.e. where can someone look to confirm this dataset has this license)

galtay (commented):
This is great. I'm pro moving forward and fixing things as we go. Lets wait and see what @jason-fries  says, but I think we can move forward with this.

bigbio/utils/license.py
```diff
@@ -0,0 +1,116 @@
...
+        string = f"{self.name}"
+
+        if hasattr(self, "version") and getattr(self, "version", None) is not None:
+            string += f" - v{self.version}"
+
+        if hasattr(self, "link") and getattr(self, "link", None) is not None:
+            string += f" ({self.link})"
+
+        if hasattr(self, "text") and getattr(self, "text", None) is not None:
+            string += f": {self.text}"
```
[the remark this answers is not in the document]
hakunanatasha: @galtay done

hakunanatasha (approved):

bigbio/utils/license.py
```diff
@@ -0,0 +1,124 @@
...
+    version: Optional[float] = None
+    provenance: Optional[str] = None
+
+    @property
+    def is_share_alike(self):
+        """
+        Is Share-alike?
+        """
+
+        return False
```
galtay, line 41: before merging this can we just raise a not implemented error here or remove?

bigbio/utils/license.py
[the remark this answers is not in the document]
galtay: @sg-wbi in the spirit of less code is easier to maintain ... dataclasses have built in string representations ... can we just use that and get rid of this function?

bigbio/utils/license.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Override to get a formatted full description of the license
+        """
+        return self.display
+
+    def __str__(self):
+        """
+        Override to get a formatted full description of the license
+        """
+        return self.display
```
galtay, line 72: I think we can get rid of these too

bigbio/utils/license.py
```diff
@@ -0,0 +1,124 @@
...
+    Optionally its version and a link to the license webpage.
+    """
+
+    def __post_init__(self):
+        if self.text is None:
+            raise ValueError(
+                "A `CustomLicense` must provide at least the license text!"
+            )
+        if self.name is None:
+            super().__setattr__("name", "CUSTOM")
```
galtay, line 89: `self.name = "CUSTOM"` ?

bigbio/utils/license.py
sg-wbi: I added a `NotImplementedError`, I like to have it there just to showcase what it is possible to do in case somebody else will work w/ this.

bigbio/utils/license.py
sg-wbi: Done

sg-wbi:
I have added `License` for all datasets, but all datasets now have a merge conflict. I tried following the instruction to resolve them and this is what git is showing me:
```python
import bigbio.utils.parsing as parsing
from bigbio.utils import schemas
from bigbio.utils.configs import BigBioConfig
<<<<<<< HEAD
from bigbio.utils.constants import Lang, Tasks
||||||| merged common ancestors
from bigbio.utils.constants import Tasks
=======
from bigbio.utils.constants import Lang, Tasks
from bigbio.utils.license import Licenses
>>>>>>> 074506192995d7011b42f4814211a63b8298f331
```
I tried modifying the import but it does not seem to help. @galtay @hakunanatasha any idea?

I could open a new clean PR but if possible I'd like to try save this one since it's a lot of work!

sg-wbi:
Closing in favor of #673

PR #632, opened 2022-05-25
Add _PUBMED metadata

sg-wbi:
This adds a `_PUBMED` module-level boolean constant specifying whether **any of the resources** used to create the corpus come form PubMed. 

This means that if there are multiple resources and at least one of them is PubMed (e.g. `mantra_gsc`) we'll have `_PUBMED = True`.

This is mostly done to facilitate pubmed id normalization (@galtay , @hakunanatasha )

By examining the resources used to build the corpora, I think at some point we would want to introduce a:

```python
@dataclass
Source:
   name : str
   link : Optional[str] = None
   description : Optional[str] = None
```
For more fine-grained retrieval, e.g. "get all copora w/ patent data".


galtay:
I think this is great as long as we communicate that this is like the language variable, i.e. if its true it just means that some portion of the dataset came from pubmed.

galtay:
I think this is great as long as we communicate that this is like the language variable, i.e. if its true it just means that some portion of the dataset came from pubmed.

galtay (dismissed):

galtay (approved):

PR #635, opened 2022-05-26
script to harmonize pubmed ids

galtay:
script to normalize pubmed and pubmed central ids


galtay:
also sneaking in a quick fix to the `diann_iber_eval` dataset (the nested iterables in the schema were causing the keywords to be blown up into characters instead of words ... also added some str.strip to help equality checks.)

PR #636, opened 2022-05-26
Fine grain tagging

hakunanatasha:
Creates a skeletal structure for fine-grained tagging using the `dataloader` config. I'm modeling this after @sg-wbi's licenses file.

Fine-grain tagging requires some discussion before we decide on the explicit class names. Using the logic of this [paper](https://arxiv.org/pdf/2112.03204.pdf), I was unable to come up with an explicit analog for biomedical literature that I felt fully comfortable with. In lieu of that, I created a [sheet](https://docs.google.com/spreadsheets/d/1l1JBgRR_l_gyUUKr9PrVgj1ALMPONQK6QpN3Ne-8BVI/edit#gid=0) that highlights a brief summary of each BB dataset's intended objective, some salient keywords, and the description provided in the dataloaders themselves.


hakunanatasha:
deprecated

PR #673, opened 2022-06-03
Normalize licenses: take 3

sg-wbi:
Same as #620, but improved.

Now even custom licenses can be accesed by `Licenses` enum.


sg-wbi:
I also modified the `test_metadata` function to ensure that the metadata have the expected type.

galtay (approved):

galtay:
@sg-wbi I posted about this PR in the slack and will merge later tonight EST (in about 5 hours) if there are no objections.

PR #685, opened 2022-06-05
Blurb gad

hakunanatasha:
Adds a BLURB config split to GAD to enable prompting

According to the BLURB dataset, fold 1 is split into a train + dev set, while the test remains the same. I included a custom blurb generator to mimic this behavior.


jason-fries (approved):
LGTM!

PR #687, opened 2022-06-06
add pubmed_metadata

sg-wbi:
Adds `_PUBMED` metadata to new datasets


galtay:
I think these are handled in other PRs, are there changes you want to keep from this PR @sg-wbi ? If not, i think we can close this.

PR #688, opened 2022-06-06
WIP Dataset Visualization Script 

ruisi-su:
this is the PR for running the stream vis app locally. `vis_app.py` is for the streamlit app, and `vis_data_card.py` is to generate the data cards. The latter file needs to be updated to generate cards for all datasets.


ruisi-su:
@jason-fries @hakunanatasha @galtay I have generated the initial batch of pdfs for:

1. datasets that are not broken
2. datasets that are not local
3. datasets that are covered by the static metadata `.json` file

Note that these are _very_ preliminary (e.g., axes names spanning through the chart). Please take a look at them and flag any changes we might need to make. Thanks!

galtay (approved):

PR #717, opened 2022-06-29
Fixes missing references in bibtex citations in all dataloaders

jason-fries:
Several `_CITATION` fields were missing the references needed to use citations in LaTeX. This fixes all such files.


galtay (approved):
:tada: :tada: :tada:

PR #723, opened 2022-07-03
change evidence-inference dataset from entailment to question answering

galtay:
The evidence-inference dataset was initially implemented as an entailment task, but this was not a great fit. This PR re-implements it as a multiple choice question answering task. 

An example from the source schema, 
```
{'id': 1,
 'prompt_id': 191,
 'pmcid': 2517154,
 'label': 'no significant difference',
 'evidence': 'Pain severity decreased during the treatment by 30% in both the placebo (P < 0.001) and infliximab groups (P < 0.001).',
 'intervention': 'Infliximab',
 'comparator': 'Placebo',
 'outcome': 'Dysmenorrhea scores'}
```

the labels come from the list ["significantly increased", "no significant difference", "significantly decreased"]. a quote from the paper, 
```
In the Evidence Inference task (Lehman et al., 2019), a model is 
provided with a full-text article describing a randomized controlled trial (RCT) 
and a ‘prompt’ that specifies an Intervention (e.g., aspirin), a Comparator 
(e.g., placebo), and an Outcome (e.g., duration of headache). 
We refer to these as ICO prompts. The task then is to infer whether a 
given article reports that the Intervention resulted in a significant increase, 
significant decrease, or produced no significant difference in the Outcome, 
as compared to the Comparator.
```

The question answering approach uses a template
```
question = (
    f"Compared to {sample['Comparator']} "
    f"what was the result of {sample['Intervention']} on {sample['Outcome']}?"
)
```

How this would appear in the new question answering format, 
```
{'id': '1',
 'question_id': '191',
 'document_id': '2517154',
 'question': 'Compared to Placebo what was the result of Infliximab on Dysmenorrhea scores?',
 'type': 'multiple_choice',
 'choices': ['significantly increased',
  'no significant difference',
  'significantly decreased'],
 'context': 'Pain severity decreased during the treatment by 30% in both the placebo (P < 0.001) and infliximab groups (P < 0.001).',
 'answer': ['no significant difference']}
```

### DIFF
--- a/bigbio/biodatasets/evidence_inference/evidence_inference.py
+++ b/bigbio/biodatasets/evidence_inference/evidence_inference.py
@@ -89,6 +89,31 @@ QA_CHOICES = [
     "significantly decreased",
 ]
 
+# Some examples are removed due to comments on the dataset's github page
+# https://github.com/jayded/evidence-inference/blob/master/annotations/README.md#caveat
+
+INCORRECT_PROMPT_IDS = set([
+    911, 912, 1262, 1261, 3044, 3248, 3111, 3620, 4308, 4490, 4491, 4324,
+    4325, 4492, 4824, 5000, 5001, 5002, 5046, 5047, 4948, 5639, 5710, 5752,
+    5775, 5782, 5841, 5843, 5861, 5862, 5863, 5964, 5965, 5966, 5975, 4807,
+    5776, 5777, 5778, 5779, 5780, 5781, 6034, 6065, 6066, 6666, 6667, 6668,
+    6669, 7040, 7042, 7944, 8590, 8605, 8606, 8639, 8640, 8745, 8747, 8749,
+    8877, 8878, 8593, 8631, 8635, 8884, 8886, 8773, 10032, 10035, 8876, 8875,
+    8885, 8917, 8921, 8118, 10885, 10886, 10887, 10888, 10889, 10890
+])
+
+QUESTIONABLE_PROMPT_IDS = set([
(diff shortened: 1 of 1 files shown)


jason-fries:
@galtay I agree the original assignment to entailment isn't correct. However, I don't think refactoring as a dataset subset is the best idea either since you're using a question template to populate the surface form of the prompt. This is what `promptsource` (and other prompting platforms) setups are designed to do, since there is a lot of choices here that can impact performance. 

That said, I don't know what the best format in BigBIO schema is for this example, since everything requires some sort of mucking with the inputs in some fashion.  Maybe something like text pairs where you have
```
 text1: "evidence: Pain severity decreased during the treatment by 30% in both the placebo (P < 0.001) and infliximab groups (P < 0.001)"
 text2: "intervention: Infliximab, comparator: Placebo, outcome: Dysmenorrhea scores"
label: no significant difference
```

This is, as you said on Slack, a great example of where prompting helps a lot, since it can give form and context to the problem. More incentive to accelerate getting prompts designed for all our tasks :)

galtay:
@jason-fries I started to implement the `text-pairs` version but realized the only task we have for `text-pairs` currently is semantic similarity so that would require introducing a new element to the `TASKS` enum https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/utils/constants.py#L56   

I think for now question-answering is an improvement over entailment and if we want to later we can re-map it into another bigbio schema (or not do that and focus on a more integrated prompt generation approach)

galtay:
we can make this a text pairs task if we want to later as the text pairs classification task was introduced here #724

Issue #731, opened 2022-07-18
Multiple normalizations per entity


phlobo:
Dear all,

I am trying to add NED annotations for different datasets (e.g., [Quaero](https://github.com/bigscience-workshop/biomedical/issues/702) and [DisTEMIST](https://github.com/bigscience-workshop/biomedical/issues/695)), and I wonder how to use the `normalized` attribute when there are multiple concept annotations for an entity in the source.

Case 1) 
For DisTEMIST, a single complex entity can contain multiple concepts, e.g.:
"adenopatías laterocervicales ipsilaterales tanto submaxilares como yugulodigástricas" is annotated with two SNOMED CT codes (127087005+127082004).

Case 2)
For many datasets like Quaero (but also MantraGSC or MedMentions), it is possible that entities have multiple normalizations, usually meaning that the mapping is ambiguous.

In my opinion, for case 1, these are actually two entities which share the same span, while for case 2 it's the same entity, so (as a user of the library) I would prefer to have a list of possible normalizations for that same entity in case 2.

This, however, is not how it is currently implemented for Mantra (https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/biodatasets/mantra_gsc/mantra_gsc.py) or MedMentions (https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/biodatasets/medmentions/medmentions.py) where each annotation becomes a new entity.

Example from Mantra:
``` 
{'id': '0084_d144.u226_T18', 'type': 'DISO', 'text': ['esterilidad'], 'offsets': [[66, 77]], 'normalized': [{'db_name': 'UMLS', 'db_id': 'C0678108'}]}
{'id': '0084_d144.u226_T21', 'type': 'DISO', 'text': ['esterilidad'], 'offsets': [[66, 77]], 'normalized': [{'db_name': 'UMLS', 'db_id': 'C0021359'}]}
```

What is the intended use of passing a list to `normalized`?

galtay:
Thanks for creating this issue and the examples @phlobo . My initial reaction is leaning towards using a single entity with multiple normalizations in all cases. @jason-fries @leonweber @shamikbose @sg-wbi what do you think here?

shamikbose:
@galtay I agree. Single entity with multiple normalizations would cover a lot of cases. How would they be uniquely identified, though? Offsets?

jason-fries:
[...]

PR #733, opened 2022-07-21
NED Annotations for Quaero (Resolves #702)

### LINKED ISSUE #702: quaero: missing NED
From the corpus webpage: https://quaerofrenchmed.limsi.fr/
```
 T1 PROC 3 16 contraception
#1 AnnotatorNotes T1 C0700589
T2 DEVI 25 50 dispositifs intra utérins
#2 AnnotatorNotes T2 C0021900
T3 ANAT 43 50 utérins
#3 AnnotatorNotes T3 C0042149 
```
The `AnnotatorNotes` are UMLS codes

phlobo:
Resolves #702 

- Annotations are read from BRAT files instead of BioC, because 1 document was missing in BioC. Numbers of documents / entities in unit test output match http://ceur-ws.org/Vol-1609/16090028.pdf
- NED annotations added
- Following discussion in https://github.com/bigscience-workshop/biomedical/issues/731, multiple CUIs per entities are currently *not* merged, as the vast majority differs by type and CUI
- Annotator Notes in BRAT can be empty, which caused an error in the parser

### DIFF
--- a/bigbio/biodatasets/quaero/quaero.py
+++ b/bigbio/biodatasets/quaero/quaero.py
@@ -211,7 +211,8 @@ class QUAERO(datasets.GeneratorBasedBuilder):
                     entity_id = f'{document_id}_{note["ref_id"]}'
                     for e in example["entities"]:
                         if e["id"] == entity_id:
-                            e["normalized"].append({"db_id": note["text"], "db_name": "UMLS"})
+                            for cui in note["text"].strip().split(" "):
+                                e["normalized"].append({"db_id": cui, "db_name": "UMLS"})
                 yield guid, example
         else:
             raise ValueError(f"Invalid config: {self.config.name}")
--- a/bigbio/biodatasets/quaero/quaero.py
+++ b/bigbio/biodatasets/quaero/quaero.py
@@ -1,13 +1,12 @@
-import os
+from pathlib import Path
 
 import datasets
-from pathlib import Path
 
 from bigbio.utils import schemas
 from bigbio.utils.configs import BigBioConfig
-from bigbio.utils.constants import Lang, Tasks, BigBioValues
+from bigbio.utils.constants import Lang, Tasks
 from bigbio.utils.license import Licenses
-from bigbio.utils.parsing import parse_brat_file, brat_parse_to_bigbio_kb
+from bigbio.utils.parsing import brat_parse_to_bigbio_kb, parse_brat_file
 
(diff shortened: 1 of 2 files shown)


bigbio/biodatasets/quaero/quaero.py
```diff
@@ -225,86 +180,39 @@ def _split_generators(self, dl_manager):
...
-                                    self._get_bioc_entity(span, doc_text)["offsets"][0]
-                                    for span in passage.annotations
-                                ],
-                            }
-                            for passage in xdoc.passages
-                        ],
-                    },
-                )
-                key += 1
+            for guid, txt_file in enumerate(folder.glob("*.txt")):
```
galtay, line 191: I'm a little concerned that `glob` will not return a deterministic order for the files. Is there a sort key on the file names that makes sense?

bigbio/biodatasets/quaero/quaero.py
```diff
@@ -225,86 +180,39 @@ def _split_generators(self, dl_manager):
...
-                for passage in xdoc.passages:
-                    for span in passage.annotations:
-                        ent = self._get_bioc_entity(span, doc_text)
-                        ent["id"] = uid  # override BioC default id
-                        data["entities"].append(ent)
-                        uid += 1
-
-                yield i, data
-                i += 1
+            for guid, txt_file in enumerate(folder.glob("*.txt")):
```
galtay, line 202: same file ordering concern here

bigbio/biodatasets/quaero/quaero.py
phlobo: The file name seems to be some kind of document id, so sorting by file name would be an option.
To be honest, I just copied that part from [examples/mlee.py](https://github.com/bigscience-workshop/biomedical/blob/master/examples/mlee.py), as this was given as a reference in the contribution guide.

>BRAT: Use [our custom brat parser](https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/utils/parsing.py). Example usage can be found in [examples/mlee.py](https://github.com/bigscience-workshop/biomedical/blob/master/examples/mlee.py).

galtay: ok, looks good, we can update in another PR if there turns out to be a canonical order that we want to enforce.  
@leonweber @sg-wbi @WojciechKusa I think we should probably update that example ^

galtay (approved):
thanks @phlobo !

PR #736, opened 2022-07-27
DisTEMIST NED Annotations (resolves #695)

### LINKED ISSUE #695: distemist: missing NED from SUPPORTED_TASKS
From the title:
"DisTEMIST corpus: detection and **normalization** of disease mentions in spanish clinical cases"
And the website, https://temu.bsc.es/distemist/:
```
DISTEMIST-linking subtrack: requires automatically finding disease mentions in published clinical cases and assigning, to each mention, a Snomed-CT term. 
```
https://github.com/bigscience-workshop/biomedical/blob/master/bigbio/biodatasets/distemist/distemist.py

phlobo:
Resolves #695

- Update DisTEMIST to final release used in shared task (5.1)
- Included NED annotations from both entity linking training sets
- Added configs for different subtracks (entities / linking)


bigbio/biodatasets/distemist/distemist.py
```diff
@@ -208,12 +214,17 @@ def _generate_examples(
...
                         entity["semantic_relations"] = row.semantic_rel.split("+")
 
                 elif self.config.schema == "bigbio_kb":
-                    entity["normalized"] = []
+                    if self.config.subset_id == "distemist_linking":
+                        entity["normalized"] = [
+                            {"db_id": code, "db_name": "SNOMED_CT"} for code in row.code.split("+")
+                        ]
+                    else:
+                        entity["normalized"] = [{"db_id": BigBioValues.NULL, "db_name": BigBioValues.NULL}]
```
galtay, line 227: I might be mixed up, but I thought we decided to go with an empty list here `entity["normalized"] = []` 
@leonweber @jason-fries ?

phlobo: I don't think there was a final conclusion on Discord. However, the empty list version causes a unit test to fail:
`Required key 'normalized' does not have any instances`

Is this acceptable?

galtay-tempus: ah, I think I remember now, this is the dataset in which some but not all configs support NED? 

if that is the case then yes, I think the unit test failing for this particular case is OK. 
currently the top level `SUPPORTED_TASKS` indicate "at least one config in the dataset supports this task" but our unit tests don't reflect that distinction at the moment. 

I think we should, 
* make the normalized entries empty lists
* ignore the unit test failure (if that's the only one) and merge this PR 
* short term - separate PR (probably for maintainers) , update the unit tests to check for "any" (as opposed to "all") of the configs matching a supported task
* longer term - separate PR (probably for maintainers), make supported tasks a config level variable/mapping

@jason-fries @leonweber @hakunanatasha

phlobo: @galtay-tempus I updated the script accordingly

galtay (approved):
sorry for the delay on this! 

added these update to the hub too https://huggingface.co/datasets/bigscience-biomedical/distemist/discussions/1

PR #745, opened 2022-10-21
add pre-commit hooks (public datasets ONLY)

ruisi-su:
I've copied the make commands to the pre-commit hooks, and this does work on public dataloaders **only** (without passing the `--data_dir` for private datasets). Not sure if it allows optional arguments, though.


hakunanatasha:
@galtay  we should consider either closing this or implementing it

PR #754, opened 2022-12-01
Optimize parse_brat_file to make fewer filesystem calls

albertvillanova:
This PR optimizes the function `parse_brat_file` by applying the ["Easier to ask for forgiveness than permission"](https://docs.python.org/3/glossary.html#term-EAFP) principle.

With this PR, for each annotation file, it will call the underlying filesystem only once (`f.open()`) instead of 2 (`f.exists()` and `f.open()`. Please note that the underlying filesystem can be either local or remote (when streaming). And remote filesystem calls have an expensive cost (network HTTP request).

If the dataset loading script has to parse 500 annotation files, with this PR it will make 500 fewer calls.

This PR also partially fixes the issue in streaming mode with all datasets that use `parse_brat_file`.

### DIFF
--- a/hub/bigbiohub.py
+++ b/hub/bigbiohub.py
@@ -307,7 +307,7 @@ def parse_brat_file(
         try:
             with annotation_file.open() as f:
                 ann_lines.extend(f.readlines())
-        except FileNotFoundError:
+        except Exception:
             continue
 
     example["text_bound_annotations"] = []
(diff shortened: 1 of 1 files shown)


albertvillanova:
@galtay could you please have a look at this PR?

Related to: https://huggingface.co/datasets/bigbio/chia/discussions/4

galtay:
sorry for delaying on this! ... can we move the changes here https://github.com/bigscience-workshop/biomedical/blob/main/bigbio/hub/bigbiohub.py

I was checking out the chia PR and was led back here. 

When we get this one merged in, I can do a global update of the `bigbiohub.py` file to all the hub repos

galtay:
@albertvillanova I made these changes here https://github.com/bigscience-workshop/biomedical/pull/846 ... we can re-open this PR if I missed something. I also uploaded the new version of bigbiohub.py to all the hub datasets.

PR #853, opened 2023-01-05
Revise implementation of BioRED

mariosaenger:
This PR improves the implementation of the BioRed corpus:

- In the previous implementation a unique entity was created per entity mention **and** database identifier. This was fixed to a single entity mention having multiple database ids.
- Furthermore, the name of the database a entity is linked to was added
- BioRed only provides abstract-level annotations for entity-linked relation pairs rather than materializing links between all surface form mentions of relation. Analogous to BC5CDR we enumerate all mention pairs concerning the entities in the triple.

### DIFF
--- a/bigbio/biodatasets/biored/biored.py
+++ b/bigbio/biodatasets/biored/biored.py
@@ -113,11 +113,11 @@ class BioredDataset(datasets.GeneratorBasedBuilder):
 
     TYPE_TO_DATABASE = {
         "CellLine": "Cellosaurus",
-        "ChemicalEntity": "Medical Subject Headings (MESH)",
-        "DiseaseOrPhenotypicFeature": "Medical Subject Headings (MESH) / Online Mendelian Inheritance in Man (OMIM)",
-        "GeneOrGeneProduct": "NCBI Gene",
-        "OrganismTaxon": "NCBI Taxonomy",
-        "SequenceVariant": "dbSNP / custom",
+        "ChemicalEntity": "MESH",
+        "DiseaseOrPhenotypicFeature": "MESH",  # Some diseases are normalized to OMIM (~ handled by special rules)
+        "GeneOrGeneProduct": "NCBIGene",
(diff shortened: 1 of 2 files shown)


galtay:
@davidkartchner can you take a look here and comment on the database names? is there a way we can make these database names consistent with the ones you are using? 

@mariosaenger is this PR sensitive to the exact db name? i.e. would it break anything if we use `MESH` instead of `"ChemicalEntity": "Medical Subject Headings (MESH)"`

davidkartchner:
@galtay @mariosaenger  In order to be consistent with other datasets, I would use the following mapping from type to database:
```
TYPE_TO_DATABASE = {
        "CellLine": "Cellosaurus",
        "ChemicalEntity": "MESH",
        "DiseaseOrPhenotypicFeature": Will need extra syntax to determine if MESH or OMIM; see [ncbi_disease](https://huggingface.co/datasets/bigbio/ncbi_disease/blob/main/ncbi_disease.py#L233)  for an example,
        "GeneOrGeneProduct": "NCBIGene",
        "OrganismTaxon": "NCBITaxon",
        "SequenceVariant": "dbSNP / custom",
    }
```

For cases where an entity type can be linked to multiple databases, it is especially important to correctly specify which database the identifier is coming from to effectively train an entity normalization model later on.  For `DiseaseOrPhenotypicFeature` or `SequenceVariant`, you can probably determine which database it links to with some basic checks on the string format (e.g. all OMIM normalization have "OMIM" prepended to their identifier).

mariosaenger:
@galtay @davidkartchner thanks for the feedback. I revised the implementation to adhere to the database naming scheme.

In a future PR, we could possibly also consider standardising the naming scheme of all NEN data sets via constants.

galtay:
thanks @mariosaenger . yes I agree, it would be nice to have a check for standardized database names in the unit tests.

galtay (approved):

PR #872, opened 2023-02-08
Closes #871

### LINKED ISSUE #871: GNormPlus: Add NLMIAT sub-part to the data set
The GNormPlus corpus consists of three sub-parts: BC2GN-Train, BC2GN-Test and NLMIAT, which are all shipped within a zip-archive. However, the last sub-part (NLMIAT) will not be read in the current implementation. This is due to the fact, that the part is also implemented as a distinct data set in Citation GIA Test Collection:

https://github.com/bigscience-workshop/biomedical/tree/main/bigbio/hub/hub_repos/citation_gia_test_collection

However, this might be contrary to the expectations of the users of the dataset and also doesn't not reflect the data discussed / used in the referenced paper. The implementation should be adapted to include all parts of the downloaded zip-archive.

mariosaenger:
The GNormPlus corpus consists of three sub-parts: BC2GN-Train, BC2GN-Test and NLMIAT, which are all shipped within a zip-archive. However, the last sub-part (NLMIAT) will not be read in the current implementation. This is due to the fact, that the part is also implemented as a distinct data set in Citation GIA Test Collection:

https://github.com/bigscience-workshop/biomedical/tree/main/bigbio/hub/hub_repos/citation_gia_test_collection

However, this might be contrary to the expectations of the users of the dataset and also doesn't not reflect the data discussed / used in the referenced paper. This PR adapts the implementation to use all parts of the downloaded zip-archive. Moreover, incorrect entity offsets are fixed.

### DIFF
--- a/bigbio/hub/hub_repos/gnormplus/gnormplus.py
+++ b/bigbio/hub/hub_repos/gnormplus/gnormplus.py
@@ -27,7 +27,7 @@ from .bigbiohub import Tasks
 from .bigbiohub import get_texts_and_offsets_from_bioc_ann
 
 
-_LANGUAGES = ['English']
+_LANGUAGES = ["English"]
 _PUBMED = True
 _LOCAL = False
 _CITATION = """\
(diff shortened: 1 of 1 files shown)


sg-wbi:
Thanks for catching this! This is indeed how the authors intended the corpus to be used (see [here](https://www.hindawi.com/journals/bmri/2015/918710/#evaluation-and-results) )

sg-wbi (approved):

PR #896, opened 2023-09-26
make czi_drsm compatible with hub

leonweber:
Makes the czi_drsm dataset compatible with hf hub by fixing the following:
- change license name to shorthand required by hf
- replace bigbio import with bigbiohub

### DIFF
--- a/bigbio/hub/hub_repos/czi_drsm/README.md
+++ b/bigbio/hub/hub_repos/czi_drsm/README.md
@@ -3,7 +3,7 @@ language:
   - en 
 bigbio_language:
   - English
-license: Creative Commons Zero v1.0 Universal
+license: cc0-1.0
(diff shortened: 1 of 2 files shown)


leonweber:
This version is already live on the hub

bigbio/hub/hub_repos/czi_drsm/README.md
```diff
@@ -15,7 +15,7 @@ bigbio_tasks:
 ---
 
 # Dataset Card for CZI DRSM
-
+[README.md](..%2Fmed_qa%2FREADME.md)
```
mariosaenger, line 18: That seems to be an mistake!?

mariosaenger (changes requested):

mariosaenger (approved):
Thanks!

PR #902, opened 2023-10-16
Make id accessible for source schema with `instance['id']`

leonweber:
Make id accessible for source schema with `instance['id']`. This is required so that we can generate a shared task submission


mariosaenger (approved):
Changes are fine.

Issue #891, opened 2023-04-13
Mantra GSC not on Hf Hub?


phlobo:
`datasets.load_dataset('bigbio/mantra_gsc')`

does not work.

There is an implementation of a loader though:
https://github.com/bigscience-workshop/biomedical/tree/main/bigbio/biodatasets/mantra_gsc

Was this implementation not migrated to the hub?

PR #923, opened 2024-06-07
Fix Mantra GSC Readme for Hugging Face Hub

phlobo:
#891 

@leonweber Fixed the readme, so the dataset can (hopefully) be pushed to the hub.


leonweber (approved):
LGTM. It is now available on the hub as `bigbio/mantra_gsc`

